Issue #50, opened 2020-09-21
Add predict api to estimator


buaazhwb:
For now, user can only obtain predictions by getting the model and then predicting

ConeyLiu:
what input and output should we want? And for predicting, does it should be the model serving part?

buaazhwb:
I'm not sure whom raydp's target user is.  Looks like the evaluate function is enough. 
If I'm a kaggler,  I need the prediction result of test dataset which is without label and submit the result to the competition. For now, I need to get the model and predict by myself.
The purpose of predicting is different from serving. We just want to obtain the prediction in a python script

Input should support both RayDp dataset and spark dataframe. For output, should it be consistent with input or just numpy array/pandas dataframe?

kira-lin:
close as stale

PR #61, opened 2020-10-27
Add predict function to estimator

buaazhwb:
#50 

- [x] Add predict function to TFEstimator

- [x] Add predict function to TorchEstimator

### DIFF
--- a/python/raydp/estimator.py
+++ b/python/raydp/estimator.py
@@ -35,7 +35,7 @@ class EstimatorInterface(ABC):
         pass
 
     @abstractmethod
-    def evaluate(self, df: PandasDataset, **kwargs) -> "numpy.array":
+    def predict(self, df: PandasDataset, **kwargs) -> "numpy.array":
         pass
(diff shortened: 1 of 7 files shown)


python/raydp/estimator.py
```diff
@@ -34,6 +34,10 @@ def fit(self, ds: PandasDataset, **kwargs) -> NoReturn:
     def evaluate(self, df: PandasDataset, **kwargs) -> NoReturn:
         pass
 
+    @abstractmethod
+    def evaluate(self, df: PandasDataset, **kwargs) -> "numpy.array":
```
ConeyLiu, line 38: `predict`?

ConeyLiu:
Hi @buaazhwb. Thanks for the `predict` API. The input of the predict is a `ParallelDataset`, so the output should also be a `ParallelDataset`. And whether should be doing the prediction with distributed mode? Because the input data can be a distributed dataset.

PR #67, opened 2020-11-26
Upgrade Ray version and use MLDataset

ConeyLiu:
- Upgrade Ray version the given master commit: `9481ecd180fb11db14394ff33bcbc8fe23ba1b51`. After this commit, we could use `MLDataset`.
- Replace `RayDataset` to `MLDataset`.


ConeyLiu:
The CI failed due to ray linux whl has not packaged java jars.

ConeyLiu:
Local MacOS test passed.

ConeyLiu:
@valiantljk, does this work for you?

ConeyLiu:
merge this, and will update the ray commit after the patch that fixes ray java jar has merged.

PR #83, opened 2021-02-04
Stop executors correctly

ConeyLiu:


core/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
```diff
@@ -132,6 +140,10 @@ class RayAppMaster(host: String,
         context.reply(success)
     }
 
+    override def onDisconnected(remoteAddress: RpcAddress): Unit = {
+      appInfo.kill(remoteAddress)
```
ConeyLiu, line 144: kill the java actor when the connection has disconnected.

PR #87, opened 2021-02-11
[python] Merge classpaths for RayDP jars with user-specified jars.

clarkzinzow:
Currently, the user is unable to specify jars that they wish to be added to the Spark driver and executor classpaths, which is a common need for e.g. [reading/writing Parquet files from/to S3](https://hadoop.apache.org/docs/r2.8.0/hadoop-aws/tools/hadoop-aws/index.html#ClassNotFoundException:_org.apache.hadoop.fs.s3a.S3AFileSystem). This PR merges any user-specified jars with the RayDP jars, so that all jars will be on the driver and executor classpaths.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,7 @@ target/
 .vscode
 
 examples/.ipynb_checkpoints/
+
+# Vim temp files
+*.swp
+*.swo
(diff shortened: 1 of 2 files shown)


carsonwang:
Thanks for the PR @clarkzinzow. Looks good! I'll merge this.

Issue #94, opened 2021-02-25
model.train() reports an error when running estimator.fit_on_spark(df)


Matrix-World:
Thanks to the author's efforts, this is a very good project. 
I tried to run the following brief test program. 
------------code-------------
import raydp
from raydp.torch import TorchEstimator
#start Spark job on Ray
spark = raydp.init_spark(
  app_name = "example test",
  num_executors = 3,
  executor_cores = 3,
  executor_memory="10GB"
)
##prepare the data
customers =  [(1,'James',21,6), (2, "Liz",25,8), (3, "John", 31, 6),\
     (4, "Jennifer", 45, 7), (5, "Robert", 41, 5), (6, "Sandra", 45, 8)]
df = spark.createDataFrame(customers, ["cID", "name", "age", "grade"])
df.show()

import torch
##create model
model = torch.nn.Sequential(torch.nn.Linear(2, 1))
optimizer = torch.optim.Adam(model.parameters())
loss = torch.nn.MSELoss()
#config
estimator = TorchEstimator(
  model = model,
  optimizer = optimizer,
  loss = loss,
  num_workers = 3,
  num_epochs = 5,
  feature_columns = ["age"],
  label_column = ["grade"]
)
print("before fit on spark, the type of model:", type(model))
estimator.fit_on_spark(df)
##get the trained model
pytorch_model = estimator.get_model()
------------code-------------
But when performing model training model.train() in training_operator.py, an error message was reported: AttributeError:'list' object has no attribute'train'.
--------error-----------
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/root/miniconda3/lib/python3.7/site-packages/raydp/torch/estimator.py", line 311, in fit_on_spark
    train_ds, evaluate_ds, num_steps, profile, reduce_results, max_retries, info)
  File "/root/miniconda3/lib/python3.7/site-packages/raydp/torch/estimator.py", line 284, in fit
    info=info)
  File "/root/miniconda3/lib/python3.7/site-packages/ray/util/sgd/torch/torch_trainer.py", line 410, in train
    num_steps=num_steps, profile=profile, info=info, dataset=dataset)
  File "/root/miniconda3/lib/python3.7/site-packages/ray/util/sgd/torch/worker_group.py", line 325, in train
    success = check_for_failure(remote_worker_stats)
  File "/root/miniconda3/lib/python3.7/site-packages/ray/util/sgd/utils.py", line 244, in check_for_failure
[...]

PR #96, opened 2021-03-02
A quick fix for issue #94

kira-lin:
In current ray versions,  models like Sequential in pytorch, which implements Iterable interface, will be treated as a list of models. A quick work-around for this is to pass in our single model as a list, then unpack it afterwards.

### DIFF
--- a/python/raydp/torch/estimator.py
+++ b/python/raydp/torch/estimator.py
@@ -209,12 +209,15 @@ class TorchEstimator(EstimatorInterface, SparkEstimatorInterface):
                 else:
                     lr_scheduler = None
 
+                # A quick work-around for https://github.com/ray-project/ray/issues/14352
+                # As we only support single model for now, this is OK
                 registered = self.register(
-                    models=model, optimizers=optimizer, criterion=loss, schedulers=lr_scheduler)
+                    models=[model], optimizers=optimizer, criterion=loss, schedulers=lr_scheduler)
                 if lr_scheduler is not None:
                     self.model, self.optimizer, self.criterion, self.scheduler = registered
                 else:
                     self.model, self.optimizer, self.criterion = registered
+                self.model = self.model[0]
 
                 # create dataset
                 batch_size = config["batch_size"]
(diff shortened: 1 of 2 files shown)


ConeyLiu:
Hi @kira-lin, could you add a test for this?

python/raydp/torch/estimator.py
```diff
@@ -209,12 +209,15 @@ def setup(self, config):
                 else:
                     lr_scheduler = None
 
+                # A quick work-around for https://github.com/ray-project/ray/issues/14352
+                # As we only support single model for now, this is OK
                 registered = self.register(
-                    models=model, optimizers=optimizer, criterion=loss, schedulers=lr_scheduler)
+                    models=[model], optimizers=optimizer, criterion=loss, schedulers=lr_scheduler)
```
ConeyLiu, line 215: should we wrap the optimizer, loss, lr_scheduler as well?

kira-lin: No need to wrap loss. I'm not sure if any optimizer implements Iterable, but I'll wrap them anyway.

kira-lin:
> Hi @kira-lin, could you add a test for this?

I modified the original test code a bit:
```python
import ray
import raydp
from raydp.torch import TorchEstimator
#start Spark job on Ray
ray.init(address='auto')
spark = raydp.init_spark(
    app_name = "example test",
    num_executors = 3,
    executor_cores = 3,
    executor_memory="10GB"
)
##prepare the data
customers = [
    (1,'James', 21, 6),
    (2, "Liz", 25, 8),
    (3, "John", 31, 6),
    (4, "Jennifer", 45, 7),
    (5, "Robert", 41, 5),
    (6, "Sandra", 45, 8)
]
df = spark.createDataFrame(customers, ["cID", "name", "age", "grade"])
df.show()

import torch
##create model
model = torch.nn.Sequential(torch.nn.Linear(1, 2), torch.nn.Linear(2,1))
optimizer = torch.optim.Adam(model.parameters())
loss = torch.nn.MSELoss()
#config
estimator = TorchEstimator(
    model = model,
    optimizer = optimizer,
    loss = loss,
    num_workers = 3,
    num_epochs = 5,
    feature_columns = ["age"],
    label_column = ["grade"],
    batch_size=2
)
print("before fit on spark, the type of model:", type(model))
estimator.fit_on_spark(df)
##get the trained model
pytorch_model = estimator.get_model()
```

ConeyLiu:
You can add it to the `tests` folder.

ConeyLiu:
thanks.

PR #99, opened 2021-03-08
add a workflow to automatically upload to pypi

kira-lin:
This workflow can upload the package to pypi whenever a tagged release is pushed to master.  Besides, we can also trigger the workflow manually. 
We @carsonwang  need to add a token to this repo before we can use it, as mentioned [here](https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/)

The nightly release is not prepared yet. Because pypi has a total size limit for each project, I think maybe we don't need to upload it nightly. Actually, building the python part is very easy, users should be able to do it. The problem is to build the java & scala part, as they have many dependencies.  Therefore, maybe we can release jar separately, and tell users how to build the package with a built jar, or modify the path to jar.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -43,3 +43,4 @@ jobs:
       uses: pypa/gh-action-pypi-publish@master
       with:
         password: ${{ secrets.PYPI_API_TOKEN }}
+        skip_existing: true
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -18,8 +18,8 @@
 name: RayDP PyPi
 
 on: 
-  push:
-    branches: [ master ]
+  schedule:
+    - cron: '0 0 * * *'
   # can manually trigger the workflow
   workflow_dispatch:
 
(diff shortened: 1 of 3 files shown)


carsonwang:
I added two secrets PYPI_API_TOKEN and TEST_PYPI_API_TOKEN. Do we need to test in test pypi?

kira-lin:
I have tested in my fork with test pypi. It should be fine.

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Build wheel
+      env:
+        RAYDP_PACKAGE_NAME: raydp_nightly
+      run: pip install wheel && ./build.sh
```
ConeyLiu, line 41: Does this work without java?

kira-lin: Oh, I forgot it. But it works in my repo, weird. Maybe it did not use java 1.8. I'll add the java part anyway.

ConeyLiu:
Maybe you need to update the `README` or update it after this merged and checked it works.

kira-lin:
Yes, please merge it so that I can check. I'll update readme afterwards.

ConeyLiu:
thanks

Issue #98, opened 2021-03-05
A bug fix in the NYC Taxi Fare Prediction example (PyTorch)


yanivg10:
To run the NYC Taxi Fare Prediction using PyTorch example, 'x.squeeze(1)' need to be changed to 'return x' (in the NYC model definition). I'm using PyTorch version 1.4.0.

ConeyLiu:
Hi @kira-lin, could you take a look?

kira-lin:
@ConeyLiu  Yes, I verified it with PyTorch version 1.7.1. With `return x.squeeze(1)` , we will have error after 30 epoches

> RayTaskError(IndexError): ray::DistributedTorchRunner.validate() (pid=18479, ip=10.0.2.139)
  File "python/ray/_raylet.pyx", line 480, in ray._raylet.execute_task
  File "python/ray/_raylet.pyx", line 432, in ray._raylet.execute_task.function_executor
  File "/home/lzhi/.pyenv/versions/raydp/lib/python3.8/site-packages/ray/util/sgd/torch/torch_runner.py", line 155, in validate
    validation_stats = self.training_operator.validate(
  File "/home/lzhi/.pyenv/versions/raydp/lib/python3.8/site-packages/ray/util/sgd/torch/training_operator.py", line 641, in validate
    metrics = self.validate_batch(batch, batch_info)
  File "/home/lzhi/.pyenv/versions/raydp/lib/python3.8/site-packages/ray/util/sgd/torch/training_operator.py", line 691, in validate_batch
    _, predicted = torch.max(output.data, 1)
IndexError: Dimension out of range (expected to be in range of [-1, 0], but got 1)

And `return x` will solve the issue.

ConeyLiu:
@kira-lin Could you help fixes it?

PR #101, opened 2021-03-10
fix issue #98

kira-lin:
Thanks @yanivg10 for this bug fix.


ConeyLiu:
thanks @kira-lin @yanivg10.

PR #102, opened 2021-03-11
fix nightly build and update readme

kira-lin:
I forgot to delete a condition in the last PR, so the scheduled run today was skipped.  Besides, pypi does not override versions,  so I change the version to be unique. We need to delete releases once we reach [project size limit](https://pypi.org/help/#project-size-limit). 
I also update the readme to include instructions for installing the nightly build.


ConeyLiu:
thanks

PR #103, opened 2021-03-12
add more examples

kira-lin:
This PR intends to add more examples to show how raydp can work with other libraries. 
You are welcome to leave a comment, if you want to see an example on a certain workload.
- [x] XGBoost
- [x] RaySGD
- [x] Horovod


examples/data_process.py
```diff
@@ -0,0 +1,91 @@
...
+            .drop("dropoff_latitude") \
+            .drop("passenger_count") \
+            .drop("key")
+    return data
+
+def nyc_taxi_preprocess(data):
+    data = clean_up(data)
+    data = add_time_features(data)
+    data = add_distance_features(data)
+    return drop_col(data)
```
ConeyLiu, line 91: add one blank line in the end.

examples/horovod_nyctaxi.py
```diff
@@ -0,0 +1,135 @@
...
+    # Start horovod workers on Ray
+    from horovod.ray import RayExecutor
+    settings = RayExecutor.create_settings(500)
+    executor = RayExecutor(settings, num_hosts=1, num_slots=2, cpus_per_slot=2)
+    executor.start()
+    # torch_ds, num_features = executor.execute_single(process_data)
+    torch_ds, num_features = process_data()
+    executor.run(train_fn, args=[torch_ds, num_features])
+    raydp.stop_spark()
+    ray.shutdown()
```
ConeyLiu, line 135: same as above

examples/pytorch_nyctaxi.py
```diff
@@ -0,0 +1,79 @@
...
+# Create a distributed estimator based on the raydp api
+estimator = TorchEstimator(num_workers=4, model=nyc_model, optimizer=optimizer, loss=criterion,
+                           feature_columns=features, label_column="fare_amount", batch_size=256,
+                           num_epochs=30)
+# Train the model
+estimator.fit_on_spark(train_df, test_df)
+# shutdown raydp and ray
+estimator.shutdown()
+raydp.stop_spark()
+ray.shutdown()
```
ConeyLiu, line 79: same as above

ConeyLiu:
Hi @kira-lin, very perfect, thanks for the awesome update. Could we also test those examples with CI too?

kira-lin:
> Hi @kira-lin, very perfect, thanks for the awesome update. Could we also test those examples with CI too?

Sure, I have updated the CI. But will github run the workflow in this commit?

.github/workflows/raydp.yml
```diff
@@ -59,3 +65,13 @@ jobs:
...
+      - name: Test Examples
+        run: |
+          ray start --head
+          python examples/random_nyctaxi.py
+          python examples/pytorch_nyctaxi.py
+          python examples/tensorflow_nyctaxi.py
+          python examples/horovod_nyctaxi.py
+          python examples/xgboost_ray_nyctaxi.py
+          python examples/raysgd_nyctaxi.py
+          ray stop
```
ConeyLiu, line 77: code style, same as the following.

ConeyLiu:
Hi @kira-lin, there are some other `jupyter` examples, do you plan to convert them as well?

kira-lin:
> Hi @kira-lin, there are some other `jupyter` examples, do you plan to convert them as well?

I think we should keep the DLRM notebook, because it's nice to have visualization in jupyter when training such large model.

For the Titanic, we already have a tensorflow nyctaxi example to shownhow to use the tensorflow estimator. And another random dataset generator is needed in case we want to run it in CI. So I think we can just keep it.

.gitignore
```diff
@@ -8,13 +8,14 @@ dist/
...
 .DS_Store
 
 .vscode
-
 examples/.ipynb_checkpoints/
 
 # Vim temp files
 *.swp
 *.swo
+*.csv
```
ConeyLiu, line 21:
```suggestion
*.csv

```

examples/data_process.py
```diff
@@ -0,0 +1,95 @@
...
+
```
ConeyLiu, line 95:
```suggestion

```

examples/horovod_nyctaxi.py
```diff
@@ -0,0 +1,136 @@
...
+    from horovod.ray import RayExecutor
+    settings = RayExecutor.create_settings(500)
+    executor = RayExecutor(settings, num_hosts=1, num_slots=1, cpus_per_slot=1)
+    executor.start()
+    # torch_ds, num_features = executor.execute_single(process_data)
+    torch_ds, num_features = process_data()
+    executor.run(train_fn, args=[torch_ds, num_features])
+    raydp.stop_spark()
+    ray.shutdown()
+
```
ConeyLiu, line 136:
```suggestion

```

examples/random_nyctaxi.py
```diff
@@ -0,0 +1,41 @@
...
+    'fare_amount':fare_amount,
+    'pickup_datetime': date,
+    'pickup_longitude': pick_long,
+    'pickup_latitude': pick_lat,
+    'dropoff_longitude': drop_long,
+    'dropoff_latitude': drop_lat,
+    'passenger_count': passenger_count
+    })
+csv_path = os.path.dirname(os.path.realpath(__file__)) + '/fake_nyctaxi.csv'
+df.to_csv(csv_path, index=False)
```
ConeyLiu, line 41:
```suggestion
df.to_csv(csv_path, index=False)

```

ConeyLiu (commented):
code style

kira-lin:
I do not know exactly why but CI seems to be fixed. Here is 2 types of errors I found in previous runs:
1. Pytorch complains about the shape. Something like:
`ValueError: Expected more than 1 value per channel when training, got input size torch.Size([1, 256])`
Once I just rerun the workflow, and it's solved. I guess it has something to do with pytorch version, like some mirror has stale version, but I'm not sure. I don't have this problem on my machine with either pytorch 1.8.1 or 1.7.1.
2. Workflow sometimes hangs because ray actor cannot be scheduled due to lack of resource. It seems like some tasks are not releasing resources, but I didn't found any. It never happened on my machine, even when I run them in a 2-cpu ray cluster just like the vm used by github. And now it's passing without significant changes to code. I guess it's because github workflow machines are too over-used. 

@ConeyLiu Please review again.

ConeyLiu:
thanks @kira-lin, LGTM

ConeyLiu:
Hi @kira-lin, it breaks the CI. Please refer to: https://discuss.pytorch.org/t/error-expected-more-than-1-value-per-channel-when-training/26274/4.

PR #108, opened 2021-03-15
only upload to pypi when the commit time is within last 24 hrs

kira-lin:
Use the timestamp of the commit on which the workflow is run, and see if it is within 24 hrs. If it is, then we go ahead and upload. Otherwise, we should have already uploaded it in the last run, so do nothing.

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -24,9 +24,21 @@ on:
   workflow_dispatch:
 
 jobs:
+  check-commit-date:
+    name: check commit date is today
+    run: |
+         :
+         timestamp=$(git log --no-walk --date=unix --format=%cd $GITHUB_SHA)
+         echo $GITHUB_SHA
+         days=$(( ( $(date --utc +%s) - $timestamp ) / 86400 ))
+         echo $days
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi.yml
```diff
@@ -39,12 +39,23 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 1.8
+    - name: days since the commit date
+      run: |
+          :
+          timestamp=$(git log --no-walk --date=unix --format=%cd $GITHUB_SHA)
+          echo $GITHUB_SHA
```
ConeyLiu, line 46: why we echo it?

.github/workflows/pypi.yml
```diff
@@ -39,12 +39,23 @@ jobs:
...
+          days=$(( ( $(date --utc +%s) - $timestamp ) / 86400 ))
+          echo $days
```
ConeyLiu, line 48: same as here

.github/workflows/pypi.yml
kira-lin: oops, forgot to delete it.

ConeyLiu:
thanks.

PR #112, opened 2021-03-23
Set Spark driver hostname to node_ip_address

ConeyLiu:


ConeyLiu:
@carsonwang, this should fix your problem.

carsonwang:
Thanks for the quick fix. This should work. I haven't tried the patch but used your idea and added the configuration in `init_spark` as a workaround and it worked. 

```
configs = {"spark.driver.host": "192.168.202.221", "spark.driver.bindAddress": "192.168.202.221"}
spark = raydp.init_spark("RayDP on K8S", num_executors=57, executor_cores=5, executor_memory="15G", configs=configs)
```

PR #133, opened 2021-04-13
External Shuffle Service

kira-lin:
This PR aims to implement external shuffle service for RayDP in order to support Dynamic Executor Allocation later. It uses Spark native External Shuffle Service for now.


kira-lin:
If we set dynamic executor allocation, executor could be killed by spark, and the object they own in ray object store might be lost. Therefore, as Ray does not support ownership transfer now, we must first write the spark data to HDFS or something else first when dynamic executor allocation is enabled. I have tested this PR with HDFS.

kira-lin:
In addition, users of Ray 1.2.0 might have issues if they have multiple NIC/IP address. This has been fixed in 1.3.0, please update ray in that case.

core/src/main/scala/org/apache/spark/deploy/raydp/RayExternalShuffleService.scala
```diff
@@ -0,0 +1,38 @@
...
+
+package org.apache.spark.deploy.raydp
+
+import io.ray.api.Ray;
+import org.apache.spark.internal.Logging
+import org.apache.spark.{SparkConf, SecurityManager}
+import org.apache.spark.deploy.ExternalShuffleService
+
+class RayExternalShuffleService() extends Logging {
+    val conf = new SparkConf()
```
ConeyLiu, line 26: The extra confs set by `raydp.init_spark` are valid here?

ConeyLiu:
Is it possible to add UT tests for this? Others, LGTM.

core/src/main/scala/org/apache/spark/deploy/raydp/RayExternalShuffleService.scala
```diff
@@ -0,0 +1,38 @@
...
+    val mgr = new SecurityManager(conf)
+    val instance = new ExternalShuffleService(conf, mgr)
+
+    def start() = {
+        instance.start()
```
carsonwang, line 31: Do we have a style checker? @ConeyLiu , did you use Spark's code style for java/scala code? Let's be consistent on the style, e.g. 2 space here as the indent.

ConeyLiu: We only have java style checker, not include scala.

core/src/main/scala/org/apache/spark/deploy/raydp/RayExternalShuffleService.scala
kira-lin: probably no. How can we get that config? Can we just put it in AppDesc?

core/src/main/scala/org/apache/spark/deploy/raydp/RayExternalShuffleService.scala
kira-lin: It seems like we have both, but in build.sh -DskipTests is set, and the checks are not performed. Actually we do not have tests for the core part.

kira-lin:
@ConeyLiu @carsonwang , I have addressed the issues, please review again. RayAppMaster now reads spark-defaults.conf, and the extra_conf provided by user is also set in RayAppMaster. An UT for external shuffle is added. I've also updated the README.

Issue #136, opened 2021-04-19
Add RayMLDataset to create MLDataset from Spark Dataframe


carsonwang:
We can add RayMLDataset to simplify how to create MLDataset from a Spark dataframe. For example: `torch_ds = RayMLDataset(spark_df, ...).to_torch(…)`
We may keep create_ml_dataset_from_spark for compatibility.

PR #137, opened 2021-04-20
Add RayMLDataset

kira-lin:
issue #136


python/raydp/spark/dataset.py
```diff
@@ -138,3 +138,18 @@ def __iter__(self):
             tb = reader.read_all()
             df: pd.DataFrame = tb.to_pandas()
             yield df
+
+class RayMLDataset(MLDataset):
+    def __init__(self, df: sql.DataFrame,
+                 num_shards: int,
+                 batch_size: int):
```
carsonwang, line 145: Is it possible to support parameter fs_directory, compression as well?

examples/xgboost_ray_nyctaxi.py
```diff
@@ -5,8 +5,9 @@
 # Please refer to https://docs.ray.io/en/latest/xgboost-ray.html to install it.
 from xgboost_ray import RayDMatrix, train, RayParams
 import raydp
+from raydp.spark.dataset import RayMLDataset
```
carsonwang, line 8: Why need import RayMLDataset twice? Can this line be removed?

python/raydp/spark/dataset.py
kira-lin: This is kind of weird to put into `__init__` because `read_parquet` directly returns an `MLDataset`. We can separate it to a static method, and change `__class__` of it to `RayMLDataset` there.

examples/xgboost_ray_nyctaxi.py
kira-lin: Removed

carsonwang:
Thanks @kira-lin , can you please fix the CI?

kira-lin:
************* Module raydp.mpi.utils
python/raydp/mpi/utils.py:44:4: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)
************* Module raydp.spark.ray_cluster_master
python/raydp/spark/ray_cluster_master.py:110:12: R1732: Consider using 'with' for resource-allocating operations (consider-using-with)

CI is failed by pylint, and is not due to my change. Maybe pylint has updated? I am actually not sure how to fix these.

PR #106, opened 2021-03-13
Switch to memory units in bytes

guykhazma:
Ray switched to memory units in bytes in this [PR](https://github.com/ray-project/ray/pull/14433)
This PR changes RayDP to use bytes unit,

PR #138, opened 2021-04-21
use ray java snapshot version

kira-lin:
This PR creates a new branch to be compatible with Ray nightly, 2.0.0-dev. The ray jar version is still 1.1.0-SNAPSHOT because they didn't update the version for long, but it's actually 2.0.0-SNAPSHOT. I have submitted a PR there to update it.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -21,7 +21,7 @@
     <fasterxml.jackson.version>2.12.2</fasterxml.jackson.version>
     <jersey.version>2.31</jersey.version>
     <scala.binary.version>2.12</scala.binary.version>
-    <ray.version>1.2.0</ray.version>
+    <ray.version>1.1.0-SNAPSHOT</ray.version>
   </properties>
 
   <repositories>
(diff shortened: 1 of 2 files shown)


ConeyLiu:
The memory unit has changed. You can refer here https://github.com/oap-project/raydp/pull/106/files

ConeyLiu:
The ray 1.3.0 has released, should we just need to support ray 1.3.0 in the master branch?

kira-lin:
I intended to have a branch which is compatible with ray-nightly. But if there is no need, we may just focus on 1.3.0.

ConeyLiu:
Hi @kira-lin, should we merge this into the master branch first?

kira-lin:
No, this PR is for 2.0.0. It's not compatible with 1.3.0, [`setJvmOptions`](https://github.com/ray-project/ray/blob/releases/1.3.0/java/api/src/main/java/io/ray/api/call/ActorCreator.java#L30) is not yet changed in 1.3.0.

PR #139, opened 2021-04-27
Bump Ray Version to 1.3.0

kira-lin:
Ray 1.3.0 Jar is ready.

### DIFF
--- a/.github/workflows/raydp.yml
+++ b/.github/workflows/raydp.yml
@@ -63,7 +63,12 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install wheel
-          pip install torch==1.8.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
+          if [ "$(uname -s)" == "Linux" ]
+          then
+            pip install torch==1.8.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
+          else
+            pip install torch
+          fi
           pip install pytest koalas tensorflow tabulate ray[tune] xgboost_ray grpcio-tools
           HOROVOD_WITH_GLOO=1
           HOROVOD_WITH_PYTORCH=1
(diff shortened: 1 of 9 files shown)


kira-lin:
It seems like Ray Tune in 1.3.0 has a [bug](https://github.com/ray-project/ray/issues/15504). In this PR I'll change the CI to bypass it.

.github/workflows/raydp.yml
```diff
@@ -63,7 +63,13 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install wheel
-          pip install pytest koalas tensorflow torch tabulate ray[tune] xgboost_ray grpcio-tools
+          if [ "$(uname -s)" == "Linux" ]
+          then
+            pip install torch==1.8.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
```
ConeyLiu, line 68: Why we need this?

core/pom.xml
```diff
@@ -21,7 +21,7 @@
     <fasterxml.jackson.version>2.12.2</fasterxml.jackson.version>
     <jersey.version>2.31</jersey.version>
     <scala.binary.version>2.12</scala.binary.version>
-    <ray.version>1.2.0</ray.version>
+    <ray.version>1.3.0</ray.version>
```
ConeyLiu, line 24: Does the package has published? I can't find it: https://mvnrepository.com/artifact/io.ray/ray-api

ConeyLiu:
Should we also upgrade the python ray dependency?

.github/workflows/raydp.yml
kira-lin: pip install torch in Linux will download ~800M, while torch-cpu is only ~100M. And we don't need gpu anyway.

core/pom.xml
kira-lin: It's published. You can check it now. I saw it [here](https://github.com/ray-project/ray/issues/15496#issuecomment-827298371)

kira-lin:
I have upgraded the required ray version. I think we can merge this PR first because it has the fix for CI. @ConeyLiu

ConeyLiu:
thanks.

PR #142, opened 2021-04-29
Revert #96 but keep the test

kira-lin:
#96 was a work-around for Ray 1.2.0 to support `torch.Sequential`. @ConeyLiu 's fix in ray is included in Ray 1.3.0, thus we no longer need this. But the unit test is kept.


ConeyLiu:
thanks.

PR #143, opened 2021-04-29
Use Spark Checkstyle

kira-lin:
Switch to the checktyle config of Apache Spark. Enable checkstyle in CI. Update the code to pass checkstyle.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -39,7 +39,7 @@ fi
 # build core part
 CORE_DIR="${CURRENT_DIR}/core"
 pushd ${CORE_DIR}
-mvn clean package -q -DskipTests
+mvn clean package -q
 popd # core dir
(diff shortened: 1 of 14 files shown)


build.sh
```diff
@@ -39,7 +39,7 @@ fi
 # build core part
 CORE_DIR="${CURRENT_DIR}/core"
 pushd ${CORE_DIR}
-mvn clean package -q -DskipTests
+mvn clean package -q
 popd # core dir
```
ConeyLiu, line 43: Could we add an option or environ parameter to disable the tests? Because we should add the UT tests for the core part, and that would be not necessary for people who want to build the project with themselves.

kira-lin: OK

PR #145, opened 2021-05-02
Fix executor class path issue

guykhazma:
The executor class path is not decoded correctly it should use `:` as a separator
(Also resolving to URL fails)


ConeyLiu:
Good catch, thanks @guykhazma. Did you meet some problems with this?

guykhazma:
@ConeyLiu I had a problem where a jar was not resolved correctly on the executer.
The fix in this PR didn't solve yet but I found this along the way issue so meanwhile I submitted this PR.
I am seeing a behaviour where a certain jar is not resolved on the executer both when using `spark.executor.extraClassPath` and `spark.jars`.
A workaround I have for now is to add the jars to pypspark jars folder and then things work fine. If I'll manage to get to it I'll try finding out why the using the parameters doesn't work as expected.

ConeyLiu:
thanks @guykhazma

PR #147, opened 2021-05-17
divide data evenly when creating MLDataset from both Spark DataFrame or Parquet files

ConeyLiu:


ConeyLiu:
This fixes the data divided unevenly problems.

PR #149, opened 2021-05-25
fix #148 enable external shuffle service when using raydp-submit

### LINKED ISSUE #148: External shuffle service cannot be used when using raydp-submit
Specifying spark.shuffle.service.enabled=true in raydp-submit will stuck.

kira-lin:
Previously when using raydp-submit, RayAppMaster did not get the correct SparkConf, and so it cannot start external shuffle service. Also, spark conf related to shuffle should be passed to external shuffle service actors. This PR solves these problems.


carsonwang:
Thanks！

PR #153, opened 2021-06-06
improve DataLoader performance when create from MLDataset

ConeyLiu:


ConeyLiu:
@kira-lin , pls take a look.

kira-lin (commented):
Thanks, LGTM

PR #156, opened 2021-06-11
limit ray to 1.3.0 in CI

kira-lin:
ray 1.4.0 has breaking change, but 1.4.0 jar is not deployed yet. So for now we limit ray to 1.3.0 in CI to pass tests


ConeyLiu:
Should we limit the ray version requirements as well?

kira-lin:
yeah, I wanted to first test whether it can pass

PR #160, opened 2021-06-18
upgrade Ray to 1.4.0

kira-lin:


kira-lin:
This looks a bit messy now.  But anyway we'll modify some of them in the remove-py4j PR, thus I'll fix it in that PR.

carsonwang:
can you also describe what was changed from Ray 1.3 -> 1.4 in the PR description that requires all these changes, and what will be fixed in your remove-py4j PR? Won't that PR require Ray nightly build?

kira-lin:
Ray 1.4.0 changed the signature of `setJvmOptions`.This function is used when java actor is created. It originally takes a `String`, now it takes a `List[String]`. When I mean messy, I mean we used to use different data structure in scala and use a `mkString` to concatenate them, but now we have to convert them to `java.util.List` using different APIs. So it just looks "messy", but no problem in function.

In the remove-py4j PR, we will need to use another way to pass options to AppMaster. I'll try to make it more tidy. `runtime_env` is under development, users might need to use ray nightly depending on their usage.

carsonwang:
Thanks! LGTM.

PR #161, opened 2021-06-21
[WIP] Remove Py4j: Use Ray cross language to start AppMaster

kira-lin:
⚠️This PR might have breaking change.

Previously, we use Py4j to start `RayAppMaster`, and it will connect to the ray cluster. In other words, it is also a driver. Therefore, the python part and java part is actually two Ray job. This might be inconvenient after object ownership transfer was introduced in Ray.

We used Py4j because using cross language call requires all python remote function/actors to reside in the code_search_path as well. In such case, users cannot define some remote functions in Jupyter notebook and play with it. However, as `runtime_env` is introduced, and [community](https://github.com/ray-project/ray/issues/12834) are trying to improve this, we decide to remove Py4j eventually. 

At the point this is merged, users will need to modify their command  to connect to ray cluster to something like this: 
```python
ray.init(address='auto', job_config=ray.job_config.JobConfig(code_search_path=raydp.get_code_search_path()))
```
If users have some python remote functions to call, they have to use `runtime_env`, or if the [issue](https://github.com/ray-project/ray/issues/12834) mentioned above is solved, hopefully no code change is needed. If users are already using cross language call, they should add their path to ours. Since `raydp.get_code_search_path` returns a list, this should be trivial.


carsonwang:
@kira-lin , it is not quite friendly to force users specifying the code_search_path when using RayDP. Instead of setting code_search_path when init Ray, is it possible to create the RayAppMaster actor using a custom runtime_env and then do cross language calls to it? This might also need changes in Ray upstream for this kind of cross language calls.

Issue #121, opened 2021-04-01
Spark 3.1.0+ not compatible


kira-lin:
Setting spark version to 3.1.0+ in `pom.xml` of core will result in this:
```
[ERROR] /home/lzhi/Projects/raydp/core/src/main/scala/org/apache/spark/scheduler/cluster/raydp/RayCoarseGrainedSchedulerBackend.scala:196: error: method doRequestTotalExecutors overrides nothing.
[ERROR] Note: the super classes of class RayCoarseGrainedSchedulerBackend contain the following, non final members named doRequestTotalExecutors:
[ERROR] protected def doRequestTotalExecutors(resourceProfileToTotalExecs: Map[org.apache.spark.resource.ResourceProfile,Int]): scala.concurrent.Future[Boolean]
[ERROR]   override protected def doRequestTotalExecutors(requestedTotal: Int): Future[Boolean] = {
```
@ConeyLiu

ConeyLiu:
It seems like spark 3.1.0 and above have changed the `CoarseGrainedSchedulerBackend` interface. We should limit the pyspark version.

kira-lin:
Yes, I've submitted a pr to revert it.

carsonwang:
If we just build our jar with Spark 3.0.0 but run it with pyspark 3.1, does it work?

kira-lin:
> If we just build our jar with Spark 3.0.0 but run it with pyspark 3.1, does it work?

It actually works. We've been doing this in CI for a few days.

PR #165, opened 2021-07-07
Flexibility in the deps

mjschock:
This assumes the external library authors are using semver.

There's still #121 to address but this unblocks those of us that want to use a newer version of pyspark.

### DIFF
--- a/.github/workflows/raydp.yml
+++ b/.github/workflows/raydp.yml
@@ -29,6 +29,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest, macos-latest ]
+        pyspark-version: [3.0.0, 3.1.2]
         python-version: [3.6, 3.7, 3.8]
 
     runs-on: ${{ matrix.os }}
@@ -51,7 +52,6 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         run: |
           sudo apt-get install -y mpich
-
       - name: Cache pip - Ubuntu
         if: matrix.os == 'ubuntu-latest'
         uses: actions/cache@v2
@@ -87,6 +87,7 @@ jobs:
         env:
           GITHUB_CI: 1
         run: |
+          pip install pyspark==${{ matrix.pyspark-version }}
           ./build.sh
           pip install dist/raydp-*.whl
       - name: Lint
--- a/.github/workflows/raydp.yml
+++ b/.github/workflows/raydp.yml
@@ -19,9 +19,9 @@ name: RayDP CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main, master ]
   pull_request:
-    branches: [ master ]
+    branches: [ main, master ]
 
 jobs:
   build-and-test:
--- a/python/setup.py
+++ b/python/setup.py
@@ -90,14 +90,14 @@ try:
     copy2(SCRIPT_PATH, SCRIPT_TARGET)
 
     install_requires = [
-        "numpy",
-        "typing",
-        "pandas == 1.1.4",
-        "psutil",
-        "pyarrow >= 0.10",
-        "ray == 1.4.0",
-        "pyspark >= 3.0.0, < 3.1.0",
-        "netifaces"
+        "numpy < 2.0.0",
+        "typing < 4.0.0",
+        "pandas >= 1.1.4, < 2.0.0",
+        "psutil < 6.0.0",
+        "pyarrow >= 0.10, < 5.0.0",
+        "ray >= 1.4.0, < 2.0.0",
+        "pyspark >= 3.0.0, < 4.0.0",
+        "netifaces < 1.0.0"
     ]
 
     _packages = find_packages()
--- a/python/setup.py
+++ b/python/setup.py
@@ -90,14 +90,14 @@ try:
     copy2(SCRIPT_PATH, SCRIPT_TARGET)
 
     install_requires = [
-        "numpy < 2.0.0",
-        "typing < 4.0.0",
-        "pandas >= 1.1.4, < 2.0.0",
-        "psutil < 6.0.0",
-        "pyarrow >= 0.10, < 5.0.0",
-        "ray >= 1.4.0, < 2.0.0",
-        "pyspark >= 3.0.0, < 4.0.0",
-        "netifaces < 1.0.0"
+        "numpy",
+        "typing",
+        "pandas >= 1.1.4",
+        "psutil",
+        "pyarrow >= 0.10",
+        "ray >= 1.4.0, < 1.5.0",
+        "pyspark >= 3.0.0",
+        "netifaces"
     ]
 
     _packages = find_packages()
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -152,7 +152,7 @@
       <plugin>
         <groupId>net.alchim31.maven</groupId>
         <artifactId>scala-maven-plugin</artifactId>
-        <version>3.2.2</version>
+        <version>3.3.3</version>
         <executions>
           <execution>
             <id>scala-compile-first</id>


.github/workflows/raydp.yml
```diff
@@ -19,9 +19,9 @@ name: RayDP CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main, master ]
```
ConeyLiu, line 22: It seems we don't have the main branch.

python/setup.py
```diff
@@ -90,14 +90,14 @@ def run(self):
...
-        "ray == 1.4.0",
-        "pyspark >= 3.0.0, < 3.1.0",
-        "netifaces"
+        "numpy < 2.0.0",
+        "typing < 4.0.0",
+        "pandas >= 1.1.4, < 2.0.0",
+        "psutil < 6.0.0",
+        "pyarrow >= 0.10, < 5.0.0",
+        "ray >= 1.4.0, < 2.0.0",
+        "pyspark >= 3.0.0, < 4.0.0",
```
ConeyLiu, line 99: Hi @kira-lin, does the raydp support spark 3.1.0?

.github/workflows/raydp.yml
mjschock: right, i added it b/c main is now the default branch rather than master.  that way the workflow runs:  https://github.com/mjschock/raydp/actions/runs/1007895832

python/setup.py
```diff
@@ -90,14 +90,14 @@ def run(self):
...
     install_requires = [
-        "numpy",
-        "typing",
-        "pandas == 1.1.4",
-        "psutil",
-        "pyarrow >= 0.10",
-        "ray == 1.4.0",
-        "pyspark >= 3.0.0, < 3.1.0",
-        "netifaces"
+        "numpy < 2.0.0",
```
ConeyLiu, line 93: Hi @mjschock, I think the changes for `pandas` and 'ray' make sense. Why you change the others? They should be flexible already.

ConeyLiu: Actually, ray has some internal changes for different versions. So we have to specify the given version.

python/setup.py
mjschock: it works fine when building with the older version in the pom.xml:  https://github.com/oap-project/raydp/issues/121#issuecomment-811790060.  need to address that issue separately but this unblocks usage of 3.1.0 which is the minimum required for, e.g., delta-spark (https://pypi.org/project/delta-spark/), which i'd like to integrate

python/setup.py
mjschock: it's a preventative measure so i don't run into things like the pyspark versioning issue in the future.  assuming other library authors play nice and abide by semver then patch and minor updates shouldn't break anything, only major.  i can reduce the changes just to pyspark if desired but i wanted to be consistant

python/setup.py
kira-lin: Well, if there are demand for spark 3.1, we should give it a try. @mjschock can you please also modify the dependencies in the raydp.yaml to see if it pass CI?
By the way, I actually have the patch to support spark 3.1, but we are in a discussion about whether we should add a shim layer for different spark versions.

mjschock: @kira-lin i updated raydp.yml to use both pyspark 3.0.0 and 3.1.2 and they both pass

python/setup.py
carsonwang: @mjschock , based on our previous experience, we usually need to make changes to support new Ray releases like 1.5. So we should probably keep `ray == 1.4.0` otherwise things may break. For most other dependencies, my understanding is setting a minimal version or without setting anything should be fine. Can we just change pyspark and pandas for now to unblock your needs?

python/setup.py
```diff
@@ -92,11 +92,11 @@ def run(self):
     install_requires = [
         "numpy",
         "typing",
-        "pandas == 1.1.4",
+        "pandas >= 1.1.4",
         "psutil",
         "pyarrow >= 0.10",
-        "ray == 1.4.0",
-        "pyspark >= 3.0.0, < 3.1.0",
+        "ray >= 1.4.0, < 1.5.0",
```
mjschock, line 98: @carsonwang  - would this be alright, allowing patch changes?

carsonwang: Yes, looks good. Thanks!

kira-lin:
It seems like this PR has gone beyond its title. Can you please explain why you merged PR #161 ? Using code search path is not ready yet, because it requires all python function/actors to be in the code search path as well. This might break our current users application.

mjschock:
> It seems like this PR has gone beyond its title. Can you please explain why you merged PR #161 ? Using code search path is not ready yet, because it requires all python function/actors to be in the code search path as well. This might break our current users application.

@kira-lin sorry about that, i should have created a separate branch.  i've done that now and reverted the changes.

i'm now running into stackoverflow issues related to ZipArchive.scala and, after a bit of research, it seemed this could be related to an old version of scala, and possibly the usage of py4j, so i pulled in the changes to see if they helped get around that issue.

kira-lin:
Oh, I see. Can you please be more specific on what causes this error? You are using Spark 3.0.0 when building the scala part, and 3.1.2 for pyspark, right? And when you call raydp.init_spark, you see this error message?

Besides, what java version are you using? It seems scala, spark and py4j versions are not modified, thus it should be working. 

Again, thanks for contributing @mjschock

mjschock:
> Oh, I see. Can you please be more specific on what causes this error? You are using Spark 3.0.0 when building the scala part, and 3.1.2 for pyspark, right? And when you call raydp.init_spark, you see this error message?
> 
> Besides, what java version are you using? It seems scala, spark and py4j versions are not modified, thus it should be working.
> 
> Again, thanks for contributing @mjschock

yes, Spark 3.0.0 in the pom.xml when i run ./build.sh.  and yes, pyspark 3.1.2 when intiializing.  java is 1.8, the openjdk version.  the example is roughly the one from the anyscale post here: https://www.anyscale.com/blog/data-processing-support-in-ray

```python
import ray
import raydp

ray.util.connect("ray-head:10001")

@ray.remote
class PySparkDriver:
  def __init__(self):
  self.spark = raydp.init_spark(
    app_name='RayDP example',
    num_executors=2,
    executor_cores=2,
    executor_memory='4GB')

  def foo(self):
    return self.spark.range(1000).repartition(10).count()

driver = PySparkDriver.remote()
print(ray.get(driver.foo.remote()))
```

mjschock:
> > Oh, I see. Can you please be more specific on what causes this error? You are using Spark 3.0.0 when building the scala part, and 3.1.2 for pyspark, right? And when you call raydp.init_spark, you see this error message?
> > Besides, what java version are you using? It seems scala, spark and py4j versions are not modified, thus it should be working.
> > Again, thanks for contributing @mjschock
> 
> yes, Spark 3.0.0 in the pom.xml when i run ./build.sh. and yes, pyspark 3.1.2 when intiializing. java is 1.8, the openjdk version. the example is roughly the one from the anyscale post here: https://www.anyscale.com/blog/data-processing-support-in-ray
> 
> ```python
> import ray
> import raydp
> 
> ray.util.connect("ray-head:10001")
> 
> @ray.remote
> class PySparkDriver:
>   def __init__(self):
>   self.spark = raydp.init_spark(
>     app_name='RayDP example',
>     num_executors=2,
>     executor_cores=2,
>     executor_memory='4GB')
> 
>   def foo(self):
>     return self.spark.range(1000).repartition(10).count()
> 
> driver = PySparkDriver.remote()
> print(ray.get(driver.foo.remote()))
> ```

anyway, i think this is unrelated.  i'll investigate further when i get a chance

kira-lin:
sorry I cannot help here since I've never seen errors like this, but I guess it's due to the environment. Did you always have the problem? Is it OK if you use pip to install our package? If so, I can upgrade the spark version now, since it'll only affect the nightly release.

kira-lin:
By the way, we are curious about in what scenario you plan to use RayDP? What feature would be useful? And if you have any other issue, feel free to ask us.

carsonwang:
Thanks @mjschock. Merged this.

PR #166, opened 2021-07-09
Add initial support for Ray dataset

kira-lin:
This PR has to use ray nightly

### DIFF
--- a/core/src/main/java/org/apache/spark/raydp/RayDPUtils.java
+++ b/core/src/main/java/org/apache/spark/raydp/RayDPUtils.java
@@ -18,6 +18,9 @@
 package org.apache.spark.raydp;
 
 import io.ray.api.ObjectRef;
+import io.ray.api.Ray;
+import io.ray.api.id.ObjectId;
+import io.ray.runtime.RayRuntimeInternal;
 import io.ray.runtime.object.ObjectRefImpl;
 
 public class RayDPUtils {
@@ -33,5 +36,17 @@ public class RayDPUtils {
     } else {
       throw new RuntimeException(obj.getClass() + " is not ObjectRefImpl");
     }
+  }
+
+  /**
+   * Create ObjectRef from Array[Byte] and register ownership.
+   * We can't import the ObjectRefImpl in scala code, so we do the conversion at here.
+   */
+  public static <T> ObjectRef<T> readBinary(byte[] obj, Class<T> clazz, byte[] ownerAddress) {
+    ObjectId id = new ObjectId(obj);
+    ObjectRefImpl<T> ref = new ObjectRefImpl<>(id, clazz);
+    ((RayRuntimeInternal) Ray.internal()).getObjectStore()
+        .registerOwnershipInfoAndResolveFuture(id, null, ownerAddress);
+    return ref;
   }
 }
--- a/python/raydp/spark/dataset.py
+++ b/python/raydp/spark/dataset.py
@@ -23,10 +23,14 @@ import pyarrow as pa
 import pyarrow.parquet as pq
 import pyspark.sql as sql
 import ray
+from ray.experimental.data.impl.block import BlockMetadata
+from ray.experimental.data.impl.block_list import BlockList
 import ray.util.data as ml_dataset
 import ray.util.iter as parallel_it
 from ray.util.data import MLDataset
 from ray.util.data.interface import _SourceShard
+from ray.experimental.data.impl.arrow_block import ArrowBlock
+from ray.experimental.data.dataset import Dataset
 
 from raydp.spark.parallel_iterator_worker import ParallelIteratorWorkerWithLen
 from raydp.utils import divide_blocks
@@ -449,3 +453,27 @@ def create_ml_dataset_from_spark(df: sql.DataFrame,
                                  node_hints: List[str] = None) -> MLDataset:
     return RayMLDataset.from_spark(
         df, num_shards, shuffle, shuffle_seed, fs_directory, compression, node_hints)
+
+@ray.remote(num_returns=2)
+def make_arrow_block(data_ref: ray.ObjectRef):
+    data = ray.get(data_ref)
+    reader = pa.ipc.open_stream(data)
+    table = reader.read_all()
+    return ArrowBlock(table), table.schema
+
+def create_ray_dataset_from_spark(df: sql.DataFrame):
+    # call java function from python
+    jvm = df.sql_ctx.sparkSession.sparkContext._jvm
+    jdf = df._jdf
+    object_store_writer = jvm.org.apache.spark.sql.raydp.ObjectStoreWriter(jdf)
+    records = object_store_writer.save()
+    blocks = []
+    block_metadatas = []
+    for record in records:
+        ref = ray.ObjectRef(record.objectId())
+        block, schema = make_arrow_block.remote(ref)
+        blocks.append(block)
+        num_records = record.numRecords()
+        block_metadatas.append(BlockMetadata(num_records, None, ray.get(schema)))
+    block_list = BlockList(blocks, block_metadatas)
+    return Dataset(block_list)
--- a/python/raydp/spark/dataset.py
+++ b/python/raydp/spark/dataset.py
@@ -21,7 +21,9 @@ import numpy as np
 import pandas as pd
 import pyarrow as pa
 import pyarrow.parquet as pq
+import pyspark
 import pyspark.sql as sql
+from pyspark.sql.types import *
 import ray
 from ray.experimental.data.impl.block import BlockMetadata
 from ray.experimental.data.impl.block_list import BlockList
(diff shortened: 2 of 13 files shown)


carsonwang:
@ericl, @wuisawesome , do you have any suggestion how to write an ArrowBlock to object store in Java? Currently what we can do is to write a `pyarrow.table` to object store in java. If we can not easily write an ArrowBlock to object store in Java, is it possible to change the definition of `ArrowBlock` to use a ref to `pyarrow.table`? So that both python and java can write a `pyarrow.table` to object store and then create an `ArrowBlock`.

ericl:
@carsonwang I filed this PR to support that, would it work for your use case? https://github.com/ray-project/ray/issues/17186

carsonwang:
@ericl , thanks a lot! That will be very helpful.

python/raydp/spark/dataset.py
```diff
@@ -449,3 +455,59 @@ def create_ml_dataset_from_spark(df: sql.DataFrame,
...
+    return DataFrame(jdf, spark._wrapped)
+
+def ray_dataset_to_spark_dataframe(spark: sql.SparkSession, ds: Dataset) -> DataFrame:
+    blocks = ds.get_blocks()
+    # assume same schema
+    schema = StructType()
+    for field in ds._blocks._metadata[0].schema:
+        schema.add(field.name, from_arrow_type(field.type), nullable=field.nullable)
+    #TODO how to branch on type of block?
+    sample = ray.get(blocks[0])
```
kira-lin, line 507: @ericl , how to branch on type of block? Maybe we can save it in metadata or blocklist class?

python/raydp/spark/ray_cluster.py
```diff
@@ -67,11 +67,12 @@ def get_spark_session(self,
...
-        driver_cp = "spark.driver.extraClassPath"
-        if driver_cp in extra_conf:
-            extra_conf[driver_cp] = ":".join(glob.glob(RAYDP_CP)) + ":" + extra_conf[driver_cp]
+        driver_cp_key = "spark.driver.extraClassPath"
+        driver_cp = ":".join(glob.glob(RAYDP_CP) + glob.glob(RAY_CP))
+        if driver_cp_key in extra_conf:
+            extra_conf[driver_cp_key] = driver_cp + ":" + extra_conf[driver_cp_key]
         else:
-            extra_conf[driver_cp] = ":".join(glob.glob(RAYDP_CP))
+            extra_conf[driver_cp_key] = driver_cp
```
kira-lin, line 75: We need to add this RAY_CP to spark's driver extra classpath because serialized owner_address is needed to register ownership in java, but we need to parse it to extract ip_address to use in spark. This might not be needed once we use cross language call to start our RayAppMaster, in that case we don't need to register ownership(?)

python/raydp/spark/dataset.py
```diff
@@ -449,3 +455,59 @@ def create_ml_dataset_from_spark(df: sql.DataFrame,
...
+    ref_list = [(ray.cloudpickle.dumps(block),) for block in blocks]
+    blocks_df = spark.createDataFrame(ref_list, schema)
+    return blocks_df.mapInPandas(_convert_blocks_to_dataframe, schema)
+
+def _convert_by_rdd(spark: sql.SparkSession, blocks: Dataset, schema: StructType) -> DataFrame:
+    object_ids = []
+    locations = []
+    for block in blocks:
+        # address is needed for locality
+        locations.append(ray.worker.global_worker.core_worker.get_owner_address(block))
```
carsonwang, line 489: Will `get_owner_address` return multiple locations if the block is stored in more than one node?

python/raydp/spark/dataset.py
```diff
@@ -449,3 +455,59 @@ def create_ml_dataset_from_spark(df: sql.DataFrame,
...
+
+def _convert_by_udf(spark: sql.SparkSession,
+                    blocks: List[ObjectRef],
+                    schema: StructType) -> DataFrame:
+    s = StructType([StructField("ref", BinaryType(), False)])
+    ref_list = [(ray.cloudpickle.dumps(block),) for block in blocks]
+    blocks_df = spark.createDataFrame(ref_list, schema)
+    return blocks_df.mapInPandas(_convert_blocks_to_dataframe, schema)
+
+def _convert_by_rdd(spark: sql.SparkSession, blocks: Dataset, schema: StructType) -> DataFrame:
```
carsonwang, line 484: Is blocks a `List[ObjectRef]` instead of `Dataset`?

python/raydp/spark/dataset.py
ericl: I think we should guarantee the block type in the dataset code, or pass in the type explicitly in the call here.

python/raydp/tests/test_spark_cluster.py
```diff
@@ -61,6 +61,12 @@ def test_spark_driver_and_executor_hostname(spark_on_ray_small):
     driver_bind_address = conf.get("spark.driver.bindAddress")
     assert node_ip_address == driver_bind_address
 
+def test_ray_dataset_from_and_to_spark(spark_on_ray_small):
+    spark = spark_on_ray_small
+    words_df = spark.createDataFrame([('look',), ('spark',), ('tutorial',), ('spark',), ('look', ), ('python', )], ['word'])
+    ds = raydp.spark.spark_dataframe_to_ray_dataset(words_df)
+    df = raydp.spark.ray_dataset_to_spark_dataframe(spark, ds)
+    assert words_df.toPandas().equals(df.toPandas())
```
ericl, line 69: Can we add a corresponding unit test on the Ray side that it isn't breaking raydp?

python/raydp/spark/dataset.py
ericl: This isn't a public API of Ray. If you really need it, you should call it from ray/data code and pass in the addresses here as an optional hint.

That way raydp isn't depending on private Ray APIs.

python/raydp/spark/dataset.py
kira-lin: Yes, I called get_blocks on dataset. Actually it should be `List[ObjectRef[Block]]`.

python/raydp/spark/dataset.py
```diff
@@ -449,3 +455,59 @@ def create_ml_dataset_from_spark(df: sql.DataFrame,
...
+        ray.init(address="auto", logging_level=logging.WARN)
+    for block in blocks:
+        dfs = []
+        for b in block["ref"]:
+            ref = ray.cloudpickle.loads(b)
+            data = ray.get(ref)
+            dfs.append(data.to_pandas())
+        yield pd.concat(dfs)
+
+def _convert_by_udf(spark: sql.SparkSession,
```
carsonwang, line 476: When converting by UDF, we also want to utilize the customized the RDD so that we can have data locality.

python/raydp/tests/test_spark_cluster.py
```diff
@@ -61,6 +61,12 @@ def test_spark_driver_and_executor_hostname(spark_on_ray_small):
     driver_bind_address = conf.get("spark.driver.bindAddress")
     assert node_ip_address == driver_bind_address
 
+def test_ray_dataset_from_and_to_spark(spark_on_ray_small):
```
carsonwang, line 64: Can you also add a test by creating a Ray Dataset directly and then convert to Spark dataframe?

python/raydp/spark/dataset.py
```diff
@@ -449,3 +455,59 @@ def create_ml_dataset_from_spark(df: sql.DataFrame,
...
+                    blocks: List[ObjectRef],
+                    schema: StructType) -> DataFrame:
+    s = StructType([StructField("ref", BinaryType(), False)])
+    ref_list = [(ray.cloudpickle.dumps(block),) for block in blocks]
+    blocks_df = spark.createDataFrame(ref_list, schema)
```
carsonwang, line 481: can we first create a RDD for ref_list here by solving the locality issue and specifying the correct partition number?

kira-lin:
The two PRs are depending on each other to pass the test, and thus is a deadlock. I have tested it in my local environment, and the newly added tests are also in the ray pr, thus will be tested there. I'm merging this now, and update later if needed.

python/raydp/spark/dataset.py
kira-lin: I have added a developer api to ray. For now addresses are needed to register ownership in java.

core/src/main/java/org/apache/spark/raydp/RayDPUtils.java
```diff
@@ -34,4 +37,16 @@ public static <T> ObjectRefImpl<T> convert(ObjectRef<T> obj) {
...
+
+  /**
+   * Create ObjectRef from Array[Byte] and register ownership.
+   * We can't import the ObjectRefImpl in scala code, so we do the conversion at here.
+   */
+  public static <T> ObjectRef<T> readBinary(byte[] obj, Class<T> clazz, byte[] ownerAddress) {
+    ObjectId id = new ObjectId(obj);
+    ObjectRefImpl<T> ref = new ObjectRefImpl<>(id, clazz);
+    ((RayRuntimeInternal) Ray.internal()).getObjectStore()
+        .registerOwnershipInfoAndResolveFuture(id, null, ownerAddress);
```
jovany-wang, line 49: `registerOwnershipInfoAndResolveFuture` is invoked in the constructor of ObjectRefImpl. I don't think we should add it any more.

PR #180, opened 2021-09-27
assign owner in from_spark so that spark executors can be killed

kira-lin:
This PR is to utilize previously implemented assigning owner in ray.put in Java. The assigned owner is now in Java, ideally it could be a python actor but not possible now. Therefore, RayAppMaster cannot be killed as long as the objects are needed. I add a stop_all, and moves stopping RayAppMaster from stop_spark to it.

Now stop_spark only stop the spark session, and is mainly for sparing resources for following workloads. Instead, stop_all will stop both the spark session and our RayAppMaster, and would be automatically called when program exits.

### DIFF
--- a/.github/workflows/raydp.yml
+++ b/.github/workflows/raydp.yml
@@ -70,32 +70,8 @@ jobs:
           SUBVERSION=$(python -c 'import sys; print(sys.version_info[1])')
           if [ "$(uname -s)" == "Linux" ]
           then
-            # install ray-nightly for linux
-            case $SUBVERSION in
-              6)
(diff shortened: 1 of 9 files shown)


kira-lin:
> Can we also add some tests?

OK.

wybryan:
Hey guys, I think there may exist easier solution for this, given that Ray 1.9.0 now has full Namespace Java API. Please check out https://github.com/oap-project/raydp/pull/198 for this.

I have provided tests named python/raydp/tests/test_data_owner_transfer.py.

Let me know your thoughts, cheers.

PR #196, opened 2021-11-10
[WIP] Workaround issues in Ray 1.8.0

carsonwang:


carsonwang:
@kira-lin , pinning the aiohttp version to 3.7.4 fixes the issue for me. But I was not able the reproduce the error in test. can you also verify the fix? I think we only need merge this to 0.4 branch.

kira-lin:
Confirmed this also fixes the issue for me.

carsonwang:
I will submit one for 0.4 branch.

PR #198, opened 2021-12-08
support data ownership transfer for conversion from Spark Dataframe to Ray Dataset 

wybryan:
1. the main code change is to support spark dataframe to ray dataset
conversion with data ownership transfer enabled

2. by-product of data ownership transfer is that raydp JVM process can
be released right after data is put into ray object store, whilst the data is
still avaiable for subsequent use without throwing "data owner has died"
exception. This is important from resource management point of view so
that resource occupied by the JVM processes are now recycled back to Ray
cluster.

3. implementaion of such feature relies on Ray 1.9.0+, in simple words,
it uses Ray namespace Java API and ray.put(data, _owner) API, and a
Singleton Object Holder actor created in python side.


wybryan:
Test cases are provided in /tests folder as well to demonstrate the functionalites.

carsonwang:
@wybryan , thanks for working on this! We will review it.

wybryan:
I refactor the code to be more cleaner :)

kira-lin:
hi @wybryan pr looks great! Can you avoid uploading the data file? You can use pandas to generate some random data, just for test.

wybryan:
@kira-lin ok，I can do that.

wybryan:
@kira-lin done.

python/raydp/tests/test_data_owner_transfer.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+
+  df_train = gen_test_data()
+
+  resource_stats = ray.available_resources()
+  cpu_cnt = resource_stats['CPU']
+
+  # convert data from spark dataframe to ray dataset,
+  # and transfer data ownership to dedicated Object Holder (Singleton)
+  df = spark_dataframe_to_ray_dataset(df_train, parallelism=4, _use_owner=True)
```
kira-lin, line 105: can you call this variable ds, to align with other tests? ds is short for dataset, df is for dataframe

python/raydp/tests/test_data_owner_transfer.py
```diff
@@ -0,0 +1,174 @@
...
+
+  # confirm that resources has been recycled
+  resource_stats = ray.available_resources()
+  assert resource_stats['CPU'] == cpu_cnt + num_executor
+
+  # confirm that data is still available from object store!
+  df.show(5)
+
+  # sanity check the dataset is as functional as normal
+  df = df.to_dask()
```
kira-lin, line 123: Can we use other function to test sanity, e.g. to_pandas? Dask is not our dependency.

python/raydp/tests/test_data_owner_transfer.py
wybryan: done

python/raydp/tests/test_data_owner_transfer.py
wybryan: done

wybryan: ds. to_pandas() will not work as tody in Ray 1.9.0, it is as expeced ,since the master head of ray-project/ray lacks support for binary input, it only supports pyarrow.Table as the underlying type of the block, as shown below (python/ray/data/impl/arrow_block.py Line 141)

    def add_block(self, block: "pyarrow.Table") -> None:
        assert isinstance(block, pyarrow.Table), block   # THIS IS A PROBLEM!
        self._tables.append(block)
        self._tables_nbytes += block.nbytes
        self._num_rows += block.num_rows

For above hassle, my simple quicky fix will be:

    def add_block(self, block: Union["pyarrow.Table", bytes]) -> None:
        if isinstance(block, bytes):
            reader = pyarrow.ipc.open_stream(block)
            tb = reader.read_all()
        else:
            tb = block
        self._tables.append(tb)
        self._tables_nbytes += tb.nbytes
        self._num_rows += tb.num_rows

To sum up, I don't think it is a raydp problem, as we don't want to serialize Java object just yet, so in my view this to_pandas() issue needs to be solved at ray project upstream. For comparison, to_pandas_ref() works perfectly.

Therefore, I'll just sanity check on dataset functionality such as mean() for our unit test :)

kira-lin:
Hi @wybryan , there are some lint errors in python, could you fix it?

kira-lin:
seems like the test fails. I think it's due to resource issue. github test instances only has 2 cores in ubuntu system. You can try to lower the number of executors and set `num_cpus` in `ray.init` in your test.

wybryan:
@kira-lin Hi Kira, the Github CI has all passed. Let me know if it is all done, cheers.

kira-lin:
Thanks again!

Issue #81, opened 2021-01-28
Add Placement Group support for Spark executor


ConeyLiu:
Support set placement group for Spark cluster.

kira-lin:
close as stale. This has been implemented

PR #200, opened 2022-01-07
Enable placement group

marin-ma:
For now, there are two ways of using placement group.
1. Specify a placement group strategy. It will create a placement group according to the strategy and the configured resources for executors. The placement group will be removed when spark session stops.
2. Specify the placement group to use, and a list of bundle indexes. If no bundle indexes is specified, all bundles will be used.

issue: https://github.com/oap-project/raydp/issues/81

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
@@ -17,23 +17,24 @@
 
 package org.apache.spark.deploy.raydp
 
+import io.ray.api.id.PlacementGroupId
+
 import java.io.File
 import java.text.SimpleDateFormat
-import java.util.{Date, Locale}
-
+import java.util.{Date, Locale, Optional}
 import scala.collection.JavaConverters._
 import scala.collection.mutable.HashMap
-
-import io.ray.api.ActorHandle
-import io.ray.api.Ray
+import io.ray.api.{ActorHandle, PlacementGroups, Ray}
+import io.ray.api.placementgroup.PlacementGroup
 import io.ray.runtime.config.RayConfig
-
 import org.apache.spark.{RayDPException, SecurityManager, SparkConf}
 import org.apache.spark.internal.Logging
 import org.apache.spark.raydp.RayExecutorUtils
 import org.apache.spark.rpc._
 import org.apache.spark.util.ShutdownHookManager
 import org.apache.spark.util.Utils
+
+import javax.xml.bind.DatatypeConverter
 
 class RayAppMaster(host: String,
                    port: Int,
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
```diff
@@ -262,6 +277,14 @@ class RayAppMaster(host: String,
...
       RayExecutorUtils.setUpExecutor(handlerOpt.get, appId, driverUrl, cores, classPathEntries)
     }
+
+    private def getNextBundleIndex: Int = {
+      if (placementGroup != null && bundleIndexes.nonEmpty) {
+        val previous = currentBundleIndex
+        currentBundleIndex = (currentBundleIndex + 1) % bundleIndexes.size
+        previous
+      } else -1
+    }
```
kira-lin, line 287: What if this is called more than expected? Is this compatible with dynamic executor allocation?

kira-lin: I guess no. Also we should update the doc and comment

marin-ma: No, this should only be a temporay solution. The placement group's python API supports using any available bundle without specifying the bundle index, but there is no such API in Java. I'm trying to upstream the Java API, and we can remove this part of code to let Ray decide which bundle to use when the API become available.

core/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
```diff
@@ -262,6 +277,14 @@ class RayAppMaster(host: String,
       val classPathEntries = appInfo.desc.command.classPathEntries.mkString(";")
       RayExecutorUtils.setUpExecutor(handlerOpt.get, appId, driverUrl, cores, classPathEntries)
     }
+
+    private def getNextBundleIndex: Int = {
+      if (placementGroup != null && bundleIndexes.nonEmpty) {
+        val previous = currentBundleIndex
+        currentBundleIndex = (currentBundleIndex + 1) % bundleIndexes.size
+        previous
+      } else -1
```
carsonwang, line 286: Let's follow Spark's code style. Unless you can write if else on the same line, otherwise always use {}. This should looks like 
```
} else {
  -1
}
```

core/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
```diff
@@ -223,6 +236,8 @@ class RayAppMaster(host: String,
         executorId, getAppMasterEndpointUrl(), cores,
         memory,
         appInfo.desc.resourceReqsPerExecutor.map(pair => (pair._1, Double.box(pair._2))).asJava,
+        placementGroup,
+        getNextBundleIndex,
```
carsonwang, line 240: What happens if an executor fails and Spark restarts a new executor?

python/raydp/tests/test_spark_cluster.py
```diff
@@ -86,5 +92,33 @@ def test_ray_dataset_to_spark(spark_on_ray_small):
     rows2 = [r.id for r in df2.take(n)]
     assert ids == rows2
 
+
+def test_placement_group(ray_cluster):
+    spark = raydp.init_spark("test_strategy", 1, 1, "500 M",
+                             placement_group_strategy="SPREAD")
```
carsonwang, line 98: Also test specifying all parameters including placement_group_strategy, placement_group with a different strategy, placement_group_bundle_indexes?

marin-ma:
I've tested placement group on the cluster. The current implementation choose the bundle to schedule executors in a round-robin manner, which is not good enough to handle some cases like executors failure. This can be improved in the future release.

Issue #199, opened 2022-01-04
Why all spark executors running on a single node in a ray cluster?


Bowen0729:
connect to a ray cluster:
ray.init(address = 'ip:6379') 

create a spark session on ray:
spark = raydp.init_spark("test",  num_executor=6, executor_memeory='5G')

and all the spark executors(ray actors) are running on single node

I hope that the executors will distributed among all nodes

what should I do?

carsonwang:
Can you try running `raydp.init_spark` in an actor and does that work? https://github.com/oap-project/raydp/issues/195#issuecomment-963732609

Bowen0729:
> Can you try running `raydp.init_spark` in an actor and does that work? [#195 (comment)](https://github.com/oap-project/raydp/issues/195#issuecomment-963732609)

Thank you for your reply.

I tried, but not worked.

I'm the first time user for ray and raydp, I start ray cluster on yarn, when I run raydp.init_spark, I saw two jobs on ray dashboard, one job contains my SparkRemote actor, the other one contains all spark executors which belong to the same node. and some of them are in 'PENDING_CREATION' state.

carsonwang:
Does your issue specific to Ray on Yarn or not? How did you run Ray on Yarn, using the existing functionality in Ray or your own implementation? Do other tasks/actors running on multiple nodes correctly?

Bowen0729:
**_How did you run Ray on Yarn, using the existing functionality in Ray or your own implementation?_**

I run Ray on Yarn via skein, and my skein config file like this:

name: ray-on-yarn
services:
    ray-head:
        instances: 1
        resources:
            vcores: 10
            memeory: 50GiB
         files:
            environment: my-conda-env.tag.gz
          script: |
              source envirnment/bin/activate
              skein kv put --key RAY_HEAD_ADDRESS --value $(hostname -i) current
              ray start --head --port 6379 --include-dashboard true --dashboard-host $(hostname -i) --block 
    ray-worker:
        instances: 2
        resources:
            vcores: 20
            memeory: 50GiB
         files:
            environment: my-conda-env.tag.gz
         script: |
              source envirnment/bin/activate
              sleep 10
              RAY_HEAD_ADDRESS=$(skein kv get --key RAY_HEAD_ADDRESS current)
              ray start --adress $RAY_HEAD_ADDRESS:6379 --block

And then skein submit to yarn, I can get a ray cluster with 3 nodes and saw them on ray dashboard

[...]

PR #201, opened 2022-01-13
expose the executor java worker classpath parameter to user

Bowen0729:
[ISSUE#199](https://github.com/oap-project/raydp/issues/199)

As for the discused in issue 199, when run ray cluster on yarn, the absolute path of the jars in each node is different, that would cause the executor started failed. Exposed the parameter will help.

### DIFF
--- a/doc/spark_on_ray.md
+++ b/doc/spark_on_ray.md
@@ -15,6 +15,16 @@ Similarly, you can also enable Dynamic Executor Allocation this way. However, cu
 ds = RayMLDataset.from_spark(..., fs_directory="hdfs://host:port/your/directory")
 ```
 
+### RayDP Executor Extra ClassPath 
+
+As raydp starts the java executors, the classpath will contain the absolute path of ray, raydp and spark by default. When you run ray cluster on yarn([Deploying on Yarn](https://docs.ray.io/en/latest/cluster/yarn.html)), the jar files are stored on HDFS, which have different absolute path in each node. In such cases,  jvm cannot find the main class and ray workers will fail to start. 
+
+To solve such problems, users can specify extra classpath when` init _spark` by configuring `raydp.executor.extraClassPath`. Make sure your jar files are distributed to the same path(s) on all nodes of the ray cluster.
+
+```python
+raydp.init_spark(..., configs={"raydp.executor.extraClassPath": "/your/extra/jar/path"})
+```
+
 ### Spark Submit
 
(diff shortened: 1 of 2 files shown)


python/raydp/spark/ray_cluster_master.py
```diff
@@ -60,6 +60,10 @@ def start_up(self, popen_kwargs=None):
 
     def _prepare_jvm_classpath(self):
         cp_list = []
+
+        if 'executor.worker.extraClassPath' in self._configs:
```
kira-lin, line 64: Please use " instead of ' to fix the lint error. Besides, I think using raydp.executor.extraClassPath is better. And we should update the doc to explain this

Bowen0729: _Please use " instead of ' to fix the lint error._ 

What does it mean? I'm a little bit confused

Bowen0729: Got it. I have changed it and added the doc

doc/spark_on_ray.md
```diff
@@ -15,6 +15,17 @@ Similarly, you can also enable Dynamic Executor Allocation this way. However, cu
 ds = RayMLDataset.from_spark(..., fs_directory="hdfs://host:port/your/directory")
 ```
 
+### RayDP Executor Extra ClassPath 
+
+As raydp starts the executor, the classpath of jvm will be the absolute path by default. When you run ray cluster on yarn([Deploying on Yarn](https://docs.ray.io/en/latest/cluster/yarn.html)), the jar files are stored on HDFS, which have different absolute path in each node. All above will cause the jvm cannot find the main class. 
+
+So, you can specify the extra classpath when init spark by configure `raydp.executo.extraClassPath`. Make sure your jar files distributed to the same path(s) on all nodes of the ray cluster.
```
kira-lin, line 22:
```suggestion
As raydp starts the java executors, the classpath will contain the absolute path of ray, raydp and spark by default. When you run ray cluster on yarn([Deploying on Yarn](https://docs.ray.io/en/latest/cluster/yarn.html)), the jar files are stored on HDFS, which have different absolute path in each node. In such cases,  jvm cannot find the main class and ray workers will fail to start. 
To solve such problems, users can specify extra classpath when` init _spark` by configuring `raydp.executor.extraClassPath`. Make sure your jar files are distributed to the same path(s) on all nodes of the ray cluster.
```

Bowen0729:
Is there anything else that needs to be modified？

PR #206, opened 2022-01-17
Update example to use Ray dataset and Ray train

KepingYan:


examples/raysgd_nyctaxi.py
```diff
@@ -61,7 +70,7 @@ def __init__(self, cols):
         self.bn4 = nn.BatchNorm1d(16)
 
     def forward(self, *x):
-        x = torch.cat(x, dim=1)
+        x = x[0]
```
kira-lin, line 73: what was the data shape? does this lose input data?

examples/raysgd_nyctaxi.py
```diff
@@ -84,42 +93,66 @@ def setup(self, config):
...
+
+def train_func(config):
+    device = torch.device(f"cuda:{train.local_rank()}" if
+                          torch.cuda.is_available() else "cpu")
+    if torch.cuda.is_available():
+        torch.cuda.set_device(device)
+
+    # Set the args to whatever values you want.
+	# If using DistributedSampler, modify it here.
+    training_operator = CustomOperator(
```
kira-lin, line 110: With ray train, we no longer need this operator, right? Let's use pure ray train(which is the new version of raysgd), check their [examples](https://docs.ray.io/en/latest/train/examples.html).

examples/raysgd_nyctaxi.py
```diff
@@ -38,12 +43,16 @@
...
+train_dataset = ray.data.from_spark(train_df, parallelism = num_executors)
+test_dataset = ray.data.from_spark(test_df, parallelism = num_executors)
 # Then convert to torch datasets
-train_dataset = train_dataset.to_torch(feature_columns=features, label_column="fare_amount")
-test_dataset = test_dataset.to_torch(feature_columns=features, label_column="fare_amount")
+feature_dtype = [torch.float] * len(features)
+train_dataset = train_dataset.to_torch(feature_columns=features, label_column="fare_amount",
+                                               label_column_dtype=torch.float, feature_column_dtypes=feature_dtype, batch_size=64)
+test_dataset = test_dataset.to_torch(feature_columns=features, label_column="fare_amount",
+                                             label_column_dtype=torch.float, feature_column_dtypes=feature_dtype, batch_size=64)
```
kira-lin, line 55: We should get shards of the dataset inside the train_func, check [here](https://docs.ray.io/en/latest/train/user_guide.html#train-datasets).

kira-lin (commented):
We should also rename it to raytrain_nyctaxi.py.

examples/raysgd_nyctaxi.py
KepingYan: I modified parameter "*x" -> "x", then the type of x is `torch.Tensor` and the shape is `[batchsize, len(features)]`.

examples/raytrain_nyctaxi.py
```diff
@@ -0,0 +1,176 @@
...
+    optimizer = torch.optim.Adam(model.parameters(), lr=lr)
+    loss_results = []
+    for epoch in range(num_epochs):
+        train_acc, train_loss = train_epoch(train_dataset, model, criterion, optimizer, device)
+        test_acc, test_loss = test_epoch(test_dataset, model, criterion, device)
+        train.report(epoch = epoch, train_acc = train_acc, train_loss = train_loss)
+        train.report(epoch = epoch, test_acc=test_acc, test_loss=test_loss)
+        loss_results.append(test_loss)
+
+trainer = Trainer(backend="torch", num_workers=num_executors, use_gpu=False)
```
kira-lin, line 151: Let num_workers be 2 to make it distributed. Claim num_cpus=4 in ray.init to pass CI.
@carsonwang Shall we include gpu support? We can't test it. This comes from a ray train example. If we support gpu, use_gpu should be a config, refer to the horovod example about how to set args.

examples/raytrain_nyctaxi.py
```diff
@@ -0,0 +1,176 @@
...
+# Set spark timezone for processing datetime
+spark.conf.set("spark.sql.session.timeZone", "UTC")
+# Transform the dataset
+data = nyc_taxi_preprocess(data)
+# Split data into train_dataset and test_dataset
+train_df, test_df = random_split(data, [0.9, 0.1], 0)
+features = [field.name for field in list(train_df.schema) if field.name != "fare_amount"]
+
+# Convert spark dataframe into ray Dataset
+train_dataset = ray.data.from_spark(train_df, parallelism = num_executors)
```
kira-lin, line 50: Remember to align parallelism with num_workers of ray train.

examples/raytrain_nyctaxi.py
```diff
@@ -0,0 +1,176 @@
...
+def train_epoch(dataset, model, criterion, optimizer, device):
+    model.train()
+    train_loss, correct, data_size = 0, 0, 0
+    for batch_idx, (inputs, targets) in enumerate(dataset):
+        inputs, targets = inputs.to(device), targets.to(device)
+        # Compute prediction error
+        outputs = model(inputs)
+        loss = criterion(outputs, targets)
+        train_loss += loss.item()
+        correct += (outputs.argmax(1) == targets).type(torch.float).sum().item()
```
kira-lin, line 96: correct is not calculated as this. This only applies to the mnist example of ray. We should compare outputs to targets to calculate accuracy.

examples/raytrain_nyctaxi.py
```diff
@@ -0,0 +1,177 @@
...
+    num_epochs = config["num_epochs"]
+    lr = config["lr"]
+    batch_size = config["batch_size"]
+    device = torch.device(f"cuda:{train.local_rank()}"
+                          if torch.cuda.is_available() else "cpu")
+    
+    # Then convert to torch datasets                
+    train_data_shard = get_dataset_shard("train")
+    train_dataset = train_data_shard.to_torch(feature_columns=features, label_column="fare_amount",
+                                               label_column_dtype=torch.float, feature_column_dtypes=feature_dtype, batch_size=batch_size)
```
kira-lin, line 134: This line might be too long

kira-lin (commented):
Almost LGTM, but it seems like we have some lint error. Can you add this examples folder to the CI lint check as well? If there are lint errors in other existing examples, fix them too.

.github/workflows/raydp.yml
```diff
@@ -94,6 +94,7 @@ jobs:
         run: |
           pip install pylint==2.8.3
           pylint --rcfile=python/pylintrc python/raydp
+          pylint --rcfile=python/pylintrc examples/raytrain_nyctaxi.py
```
kira-lin, line 97: make it examples/*.py?

KepingYan: Ok, I'll push again after modifying other files.

carsonwang:
@KepingYan , are those formatting changes automatically from your IDE settings? If you do want to update them, we'd better to do them in a separate PR.

KepingYan:
changes based on previous test results:

- transfer the train, test dataset split operation on ray dataset.

- during dataframe->ray data, use ‘repartition()’ to perform resharding operation.

- switch evaluation metric from accuracy to MSE. The effect can be seen on source dataset, and is not obvious on randomly obtained dataset.

examples/raytrain_nyctaxi.py
```diff
@@ -0,0 +1,168 @@
...
+data = spark.read.format("csv").option("header", "true") \
+        .option("inferSchema", "true") \
+        .load(NYC_TRAIN_CSV)
+# Set spark timezone for processing datetime
+spark.conf.set("spark.sql.session.timeZone", "UTC")
+# Transform the dataset
+data = nyc_taxi_preprocess(data)
+# Convert spark dataframe into ray Dataset
+# pylint: disable=unbalanced-tuple-unpacking
+dataset = ray.data.from_spark(data, parallelism = num_executors)
```
carsonwang, line 43: Remove parallelism here?

KepingYan: Done.

PR #214, opened 2022-02-22
Test raydp for multiple versions of ray

kira-lin:


ConeyLiu:
Just a suggestion. In production, the software is updated much more slowly than the upstream update/release. It would be better to support more versions of basic services in our product.

carsonwang:
Thanks @ConeyLiu for the suggestion. Probably we can test multiple Ray versions in the test and make sure they can work.

PR #216, opened 2022-02-24
reorg as RayDPMaster

kira-lin:
This PR refactor the spark master class. Now it is an ray actor, and might not start on the node where init_spark is called. Spark driver will be on the local node, and connects to the remote spark cluster. This way is more convenient when working with ray client. 
Meanwhile, some stale code is deleted.


python/raydp/services.py
```diff
@@ -1,109 +0,0 @@
...
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
-
-from abc import ABC, abstractmethod
-from typing import Any, Dict, NoReturn
-
-
-class Cluster(ABC):
```
carsonwang, line 22: I believe this is for supporting multiple big data frameworks in RayDP, not just for Spark. There are still possibility for that. So should we keep this? cc @ConeyLiu

ConeyLiu: That's right. I think we should keep it left if we want to support more distributed frameworks. Some other frameworks (such as flink, presto) are all popular and widely used in many companies.

PR #219, opened 2022-03-03
make raydp work with the latest Ray

KepingYan:
The latest Ray does not enable redis by default, but uses gcs bootstrap mode. So gcs_address is used to set "ray.address", and the code takes into account the previous version.

### DIFF
--- a/python/raydp/spark/ray_cluster_master.py
+++ b/python/raydp/spark/ray_cluster_master.py
@@ -152,8 +152,6 @@ class RayClusterMaster(ClusterMaster):
 
         options["ray.run-mode"] = "CLUSTER"
         options["ray.node-ip"] = node.node_ip_address
-        options["ray.address"] = node.redis_address
-        options["ray.redis.password"] = node.redis_password
         options["ray.logging.dir"] = node.get_logs_dir_path()
         options["ray.session-dir"] = node.get_session_dir_path()
         options["ray.raylet.node-manager-port"] = node.node_manager_port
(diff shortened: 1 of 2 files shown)


python/raydp/spark/ray_cluster_master.py
```diff
@@ -194,3 +199,9 @@ def stop(self):
             self._gateway = None
 
         self._started_up = False
+
+    def get_gcs_status(self):
+        try:
+            return ray._private.gcs_utils.use_gcs_for_bootstrap()
```
carsonwang, line 205: It is private so probably we should not use this API. Is there a public interface we can use to check if it is using gcs for bootstrap?

KepingYan: I haven't found a more suitable one. Maybe we could implement this function ourselves:  https://github.com/ray-project/ray/blob/991a62dd4782e17a519ce29495b6e837f37f19f3/python/ray/_private/gcs_utils.py#L98-L103

carsonwang: In latest ray people can still change the configuration to start local redis. Can we directly use `Config.bootstrap_with_gcs()`?

kira-lin:
```python
@property
    def address(self):
        """Get the address for bootstrapping, e.g. the address to pass to
        `ray start` or `ray.int()` to start worker nodes, that has been
        converted to ip:port format.
        """
        if use_gcs_for_bootstrap():
            return self._gcs_address
        return self._redis_address
```
I think ray already handle the case?

KepingYan: So could we directly use node.address? But I think we also need to know ray's version and decide whether to set ray.redis.password or not. In [pr](https://github.com/ray-project/ray/pull/22788), I think ray.redis.password may not be available in the future. (btw: I haven't fully understood this pr)

kira-lin: We need to know ray's version because we plan to support multiple ray versions. In previous versions, we have to use redis_address.

iycheng:
Could you try to update this line to use `node.address`?
https://github.com/oap-project/raydp/blob/c08a7867708224c25a73ae326d9244492b3f71fd/python/raydp/spark/ray_cluster_master.py#L146

KepingYan:
> Could you try to update this line to use `node.address`?
> 
> https://github.com/oap-project/raydp/blob/c08a7867708224c25a73ae326d9244492b3f71fd/python/raydp/spark/ray_cluster_master.py#L146

Thanks, it's updated.

PR #227, opened 2022-04-08
fix the issue: raydp-0.4.1 does not work with ray 1.11.0

kira-lin:
apply the update: node.redisaddress -> node.address


carsonwang:
LGTM

carsonwang:
LGTM

PR #233, opened 2022-04-20
migrate TorchEstimator to use Ray Train API and Ray Dataset

KepingYan:
**Parameters change details:**
- \+ resources_per_worker: used in Trainer. The resources reserved for each worker.
- \+ drop_last: used in RayDataset.to_torch(). Determine whether to drop the last incomplete batch.
- \+ callbacks: used in Trainer.run(). User can customize functions that can be executed during training.
- \- scheduler_step_freq: determine when scheduler.step is called. 
　　"batch": step will be called after every optimizer step.
　　"epoch": step will be called after one pass of the DataLoader.
　　"manual":  the scheduler will not be incremented automatically.
　　Now the code is the 'batch' mode, and this can meet basic needs. So this can be optimized as a trick for model training.
- \- feature_shapes & label_shape: required in MLDataset.to_torch(). They're used to check data. In RayDataset.to_torch(), data is checked by comparing      feature_columns and feature_column_dtypes.
- \- shuffle:  https://docs.ray.io/en/master/train/user_guide.html#example-per-epoch-shuffle-pipeline (provided by kira-lin)
- \- fs_directory & compression: Required in RayMLDataset.from_spark().
- \- num_steps & profile=False & reduce_results & info: Required in TorchTrainer.train().

**Can be optimized:**
- User are able to provide their train_fun().
- Multi-task training can be satisfied.
- `scheduler_step_freq` can also be supported.

### DIFF
--- a/python/raydp/torch/estimator.py
+++ b/python/raydp/torch/estimator.py
@@ -107,7 +107,7 @@ class TorchEstimator(EstimatorInterface, SparkEstimatorInterface):
         :param drop_last: Set to True to drop the last incomplete batch
         :param num_epochs: the total number of epochs will be train
         :param num_processes_for_data_loader: the number of processes use to speed up data loading
-        :param callbacks: which will be executed during training. 
+        :param callbacks: which will be executed during training.
         :param extra_config: the extra config will be set to ray.train.Trainer
         """
         self._num_workers = num_workers
@@ -194,7 +194,8 @@ class TorchEstimator(EstimatorInterface, SparkEstimatorInterface):
                                                 drop_last=config["drop_last"])
         if config["evaluate"]:
             evaluate_data_shard = get_dataset_shard("evaluate")
-            evaluate_dataset = evaluate_data_shard.to_torch(feature_columns=config["feature_columns"],
+            evaluate_dataset = evaluate_data_shard.to_torch(
+                                                    feature_columns=config["feature_columns"],
                                                     label_column=config["label_column"],
                                                     label_column_dtype=config["label_type"],
                                                     feature_column_dtypes=config["feature_types"],
--- a/python/raydp/torch/estimator.py
+++ b/python/raydp/torch/estimator.py
@@ -16,25 +16,95 @@
 #
 
 import inspect
-from typing import Any, Callable, List, NoReturn, Optional, Union
+from typing import Any, Callable, List, NoReturn, Optional, Union, Dict
+from numpy import outer
 
 import torch
-from ray.util.data import MLDataset
-from ray.util.sgd.torch.torch_dataset import TorchMLDataset
-from ray.util.sgd.torch.torch_trainer import TorchTrainer
-from ray.util.sgd.torch.training_operator import TrainingOperator
 from torch.nn.modules.loss import _Loss as TLoss
-from torch.utils.data.dataloader import DataLoader
 
 from raydp.estimator import EstimatorInterface
-from raydp.spark import RayMLDataset
 from raydp.spark.interfaces import SparkEstimatorInterface, DF, OPTIONAL_DF
 
+from ray import train
+from ray.train import Trainer, TrainingCallback, get_dataset_shard
+import ray.data
+from ray.data.dataset import Dataset
+from ray.data.impl.arrow_block import ArrowRow
+
+def train_fun(config):
+    # create model
+    if isinstance(config["model"], torch.nn.Module):
+        model = config["model"]
+    elif callable(config["model"]):
+        model = config["model"](config)
+    else:
+        raise Exception(
+            "Unsupported parameter, we only support torch.nn.Model instance "
+            "or a function(dict -> model)")
+
+    # create optimizer
+    if isinstance(config["optimizer"], torch.optim.Optimizer):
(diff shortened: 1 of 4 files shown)


python/raydp/torch/estimator.py
```diff
@@ -16,25 +16,98 @@
...
+            "Unsupported parameter, we only support torch.nn.Model instance "
+            "or a function(dict -> model)")
+
+    # create optimizer
+    if isinstance(config["optimizer"], torch.optim.Optimizer):
+        # it is the instance of torch.optim.Optimizer subclass instance
+        # rewrite the optimizer
+        optimizer_cls = config["optimizer"].__class__
+        state = config["optimizer"].state_dict()
+        optimizer = optimizer_cls(model.parameters(), lr=0.1)  # lr must pass for SGD
```
kira-lin, line 51: why is this 0.1 fixed?

python/raydp/torch/estimator.py
```diff
@@ -156,154 +209,93 @@ def _check(self):
...
-
-    def _create_tf_ds(self, ds: MLDataset) -> TorchMLDataset:
-        return ds.to_torch(self._feature_columns,
-                           self._feature_shapes,
-                           self._feature_types,
-                           self._label_column,
-                           self._label_shape,
-                           self._label_type)
+    @staticmethod
+    def train_epoch(dataset, model, criterion, optimizer, scheduler=None):
```
kira-lin, line 213: Is it necessary to be a static method?

python/raydp/torch/estimator.py
```diff
@@ -156,154 +209,93 @@ def _check(self):
...
+            optimizer.step()
+            if scheduler is not None:
+                scheduler.step()
+
+        train_loss /= (batch_idx + 1)
+        train_acc = correct/data_size
+        return train_acc, train_loss
+
+    @staticmethod
+    def evaluate_epoch(dataset, model, criterion):
```
kira-lin, line 236: similarly.

python/raydp/torch/estimator.py
```diff
@@ -16,25 +16,98 @@
...
-from raydp.spark import RayMLDataset
 from raydp.spark.interfaces import SparkEstimatorInterface, DF, OPTIONAL_DF
 
+from ray import train
+from ray.train import Trainer, TrainingCallback, get_dataset_shard
+import ray.data
+from ray.data.dataset import Dataset
+from ray.data.impl.arrow_block import ArrowRow
+
+def train_fun(config):
```
kira-lin, line 34: Is this configurable now? Maybe we should make this function configurable for some advanced use, what do you think?

python/raydp/torch/estimator.py
KepingYan: I have tested this. The `lr` here is a default value, which is used when the optimizer passed in has no `lr`, otherwise it will be updated to the value set by user.

python/raydp/torch/estimator.py
KepingYan: I think it could be called by class.func() without creating a class object.

python/raydp/torch/estimator.py
kira-lin: How can user set the lr value, is it included in model.parameters()?

python/raydp/torch/estimator.py
kira-lin: I felt a little weird because train_fun is defined out of the class, while these two functions are static methods of the class. Why is it so?

python/raydp/torch/estimator.py
kira-lin: The behavior is same as before, so it's ok. I just thought it would be good if users are able to provide their train_fun, when the default one is not enough. For example, if they want to train a GAN network, then this is not enough, right?

python/raydp/torch/estimator.py
KepingYan: lr is included in the optimizer passed in by user, and will be updated in Line52 through `state`

python/raydp/torch/estimator.py
KepingYan: I was troubled by this. I want to define train_fun() in the class, but the Trainer would not find it. Maybe I should put train_epoch() and evalue_epoch() together out of the class?

kira-lin: Yes, I think that's better, if train_fun cannot be in the class.

KepingYan: > I was troubled by this. I want to define train_fun() in the class, but the Trainer would not find it. Maybe I should put train_epoch() and evalue_epoch() together out of the class?

hmmm I retested this, it works.

kira-lin:
Can you fix the test? why does torch sequential fail while torch succeeds?

KepingYan:
@kira-lin it‘s because torch sequential triggers this [problem](https://github.com/ray-project/ray/pull/23398), the ray version used by CI hasn’t solved it. Maybe I should avoid failure by modifying the data sample?

kira-lin:
@KepingYan OK, so there should be no problem if we merge this?

KepingYan:
@kira-lin  The previous failure was avoided by modifying the data sample, now the CI passes.

PR #238, opened 2022-05-20
Provide metric options for TorchEstimator

KepingYan:


python/raydp/torch/torch_metrics.py
```diff
@@ -0,0 +1,44 @@
+import torchmetrics
+import sys
+module = sys.modules[__name__]
+
+class Torch_Metric():
```
carsonwang, line 5: Change to `TorchMetric` will be more consistent with the style in the project.

python/raydp/torch/torch_metrics.py
```diff
@@ -0,0 +1,44 @@
...
+    def __init__(self, metrics_name, metrics_config):
+        self._metrics_name = metrics_name
+        self._preprocess_fun = {}
```
carsonwang, line 8: What about rename `fun` to `func` in all places?

carsonwang (commented):
What if torchmetrics is not installed? Is there any messages that remind the user to install or we add something to document?

python/raydp/torch/torch_metrics.py
KepingYan: done.

python/raydp/torch/torch_metrics.py
KepingYan: thanks. done.

KepingYan:
> What if torchmetrics is not installed? Is there any messages that remind the user to install or we add something to document?

if I use try...except... to import torchmetrics, it'll triggers pylint `import-outside-toplevel`, should I set to ignore this in code?

python/raydp/torch/estimator.py
```diff
@@ -119,6 +122,14 @@ def __init__(self,
                Note: Now the value can only be False
         :param num_processes_for_data_loader: the number of processes use to speed up data loading
         :param callbacks: which will be executed during training.
+        :param metrics_name: the name of metrics' classes used to evaluate model. the name can be
+               used refer: https://torchmetrics.readthedocs.io/en/latest/, For example:
```
carsonwang, line 126: What about changing "the name can be used refer:" to "The full name list is available at:".

python/raydp/torch/estimator.py
```diff
@@ -119,6 +122,14 @@ def __init__(self,
...
+               for classification tasks, it can be "Accuracy", "Precision" and "Recall";
+               for regression tasks, it can be "MeanAbsoluteError" or "MeanSquaredError".
+               You can also pass a custom torchmetrics.Metric, the class should be defined
+               with reference to https://torchmetrics.readthedocs.io/en/latest/pages/implement.html.
+        :param metrics_config: the config can be set to metrics which you need to set. Its format
```
carsonwang, line 131: Maybe simplify "the config can be set to metrics which you need to set." to something like "the optional config for the metrics."

python/raydp/torch/torch_metrics.py
```diff
@@ -0,0 +1,61 @@
...
+            assert isinstance(metrics_name, list), "metrics_name must be a list"
+            for metric in self._metrics_name:
+                if isinstance(metric, torchmetrics.Metric):
+                    self._preprocess_func[metric.__class__.__name__] = None
+                    self._metrics_func[metric.__class__.__name__] = metric
+                elif isinstance(metric, str) and hasattr(torchmetrics, metric):
+                    self._preprocess_func[metric] = getattr(module, "pre"+metric, None)
+                    if metrics_config is not None and metrics_config[metric] is not None:
+                        self._metrics_func[metric] = getattr(torchmetrics, metric)(
+                                                            **metrics_config[metric])
```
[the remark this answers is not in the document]
KepingYan: It is the parameter corresponding to a concrete metric class of torchmetrics ( I added this to the comments). So user can refer to torchmetrics doc.

python/raydp/torch/torch_metrics.py
```diff
@@ -0,0 +1,61 @@
...
+        self._preprocess_func = {}
+        self._metrics_func = {}
+        if self._metrics_name is not None:
+            assert isinstance(metrics_name, list), "metrics_name must be a list"
+            for metric in self._metrics_name:
+                if isinstance(metric, torchmetrics.Metric):
+                    self._preprocess_func[metric.__class__.__name__] = None
+                    self._metrics_func[metric.__class__.__name__] = metric
+                elif isinstance(metric, str) and hasattr(torchmetrics, metric):
+                    self._preprocess_func[metric] = getattr(module, "pre"+metric, None)
```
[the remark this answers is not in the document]
KepingYan: The purpose of _preprocess_func is: if we need some operations before calling the method of torchmetrics, we can define a _preprocess_func for a torchmetrics method. And there is no real implementation now.

python/raydp/torch/estimator.py
KepingYan: okay

carsonwang:
LGTM

PR #241, opened 2022-05-24
Make Ray MLDataset an optional dependency

jjyao:
Ray 2.0 will remove deprecated MLDataset, which will break raydp. This PR makes MLDataset an optional dependency so raydp will work with Ray 2.0.  Eventually we should delete MLDataset from raydp and use Ray Dataset instead.

### DIFF
--- a/python/raydp/spark/__init__.py
+++ b/python/raydp/spark/__init__.py
@@ -15,16 +15,22 @@
 # limitations under the License.
 #
 
-from .dataset import RayMLDataset, \
-                     spark_dataframe_to_ray_dataset, \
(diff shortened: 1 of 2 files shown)


kira-lin:
Have you tested this solved the issue in ray? In estimator.py, torch/, tf/, MLDataset is also imported. I think we also need to change that.

jjyao:
In the Ray test (i.e test_raydp_dataset.py), estimator.py, torch, tf, etc are not imported so it should be fine. Also I think you will change them to use Ray Dataset instead so I didn't touch them.

kira-lin:
OK, I'll merge this first, and release the nightly soon

PR #243, opened 2022-05-31
Migrate TFEstimator to ray train

KepingYan:


python/raydp/tf/estimator.py
```diff
@@ -75,168 +77,200 @@ def __init__(self,
...
-            assert isinstance(metrics, list), "metrics must be a list"
-            if isinstance(metrics[0], str):
-                _metrics = metrics
-            elif isinstance(metrics[0], keras.metrics.Metric):
-                _metrics = [keras.metrics.serialize(m) for m in metrics]
+            assert isinstance(config["metrics"], list), "metrics must be a list"
+            if isinstance(config["metrics"][0], str):
+                serialized_metrics = config["metrics"]
+            elif isinstance(config["metrics"][0], keras.metrics.Metric):
+                serialized_metrics = [keras.metrics.serialize(m) for m in config["metrics"]]
```
carsonwang, line 148: Why do we need to serialize and deserialize the metrics in the same method?

python/raydp/tf/estimator.py
```diff
@@ -75,168 +77,200 @@ def __init__(self,
...
         # TODO: we should support multiple losses for multiple outputs model
-        assert loss is not None, "loss must not be None"
-        if isinstance(loss, str):
-            _loss = loss
-        elif isinstance(loss, keras.losses.Loss):
-            _loss = keras.losses.serialize(loss)
+        if isinstance(config["loss"], str):
+            serialized_loss = config["loss"]
+        elif isinstance(config["loss"], keras.losses.Loss):
+            serialized_loss = keras.losses.serialize(config["loss"])
```
carsonwang, line 133: Why do we need to serialize and deserialize the loss in the same method?

python/raydp/tf/estimator.py
```diff
@@ -75,168 +77,200 @@ def __init__(self,
...
-            if "config" in self._extra_config:
-                self._extra_config["config"].update(config)
-            else:
-                self._extra_config["config"] = config
-        else:
-            self._extra_config = {"config": config}
-
-        self._num_epochs: int = num_epochs
+    @staticmethod
+    def train_fun(config):
```
carsonwang, line 159: Rename to `train_func`

python/raydp/tf/estimator.py
```diff
@@ -75,168 +77,200 @@ def __init__(self,
...
+        )
+        return results
 
     def fit_on_spark(self,
                      train_df: DF,
                      evaluate_df: OPTIONAL_DF = None,
-                     fs_directory: Optional[str] = None,
-                     compression: Optional[str] = None) -> NoReturn:
+                     max_retries=3,
+                     stop_spark_after_conversion=False) -> NoReturn:
```
carsonwang, line 247: Can we still support `fs_directory` and `compression` parameters so users can persist the data to disk in some cases? Let's keep the parameters but we can implement it in a followup PR.

python/raydp/tf/estimator.py
```diff
@@ -15,25 +15,26 @@
...
-from raydp.spark import RayMLDataset
 from raydp.spark.interfaces import SparkEstimatorInterface, DF, OPTIONAL_DF
-
+from raydp import stop_spark
+from raydp.spark import spark_dataframe_to_ray_dataset
 
 class TFEstimator(EstimatorInterface, SparkEstimatorInterface):
     def __init__(self,
                  num_workers: int = 1,
+                 parallelism: int = None,
```
carsonwang, line 37: Perhaps we don't need to expose the parallelism parameter which seems to expose too much details. The default Spark parallelism works fine in our tests. Otherwise a user can always repartition the Spark DF to get the expected parallelism. Can we remove this parameter?

python/raydp/tf/estimator.py
```diff
@@ -75,168 +77,200 @@ def __init__(self,
...
-                evaluate_df, self._num_workers, self._shuffle, None, fs_directory, compression)
-        return self.fit(train_ds, evaluate_ds)
+            evaluate_df = self._check_and_convert(evaluate_df)
+            evaluate_ds = spark_dataframe_to_ray_dataset(evaluate_df,
+                                                         parallelism=self._parallelism,
+                                                         _use_owner=stop_spark_after_conversion)
+        if self._parallelism is None:
+            num_blocks = self._num_workers * 4
+            # TODO: we should support shuffle=self._shuffle when ray updates (ray issue#26059)
+            train_ds = train_ds.repartition(num_blocks=num_blocks, shuffle=True)
```
carsonwang, line 262: Let's remove repartition and let Ray train to handle it.

carsonwang (commented):
Thanks @KepingYan for the work!

python/raydp/tf/estimator.py
KepingYan: Sry, I didn't realize this problem before, and I corrected it as previous method. Thx

python/raydp/tf/estimator.py
KepingYan: done

python/raydp/tf/estimator.py
KepingYan: Done.

carsonwang:
Thanks @KepingYan. I'll merge this. The failed CI is unrelated to this PR.

Issue #250, opened 2022-07-03
Support custom installed Spark from $SPARK_HOME


pang-wu:
Currently the Spark classpath is inferred by the pyspark install location. In our environment I need to support a custom installed Spark location because I can't control/know where pyspark is installed (and might not have access as well).

carsonwang:
Fixed by #251

PR #251, opened 2022-07-03
Support custom Spark location by picking up $SPARK_HOME

pang-wu:
This should address issue: https://github.com/oap-project/raydp/issues/250

The failed check will be fixed by [this PR](https://github.com/oap-project/raydp/pull/248).

### DIFF
--- a/.github/workflows/raydp.yml
+++ b/.github/workflows/raydp.yml
@@ -76,7 +76,7 @@ jobs:
           else
             pip install torch
           fi
-          pip install ray==1.12 pytest koalas tensorflow==2.5.1 tabulate xgboost_ray[default] grpcio-tools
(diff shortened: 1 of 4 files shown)


carsonwang:
Thanks @pang-wu . The change looks good. I'll see if we can fix the CI first and run the added test.

pang-wu:
Thank you @carsonwang , one more thing though.. I can't seems to find a way to confirm in the test Spark is actually using the dir from `SPARK_HOME` rather than the on installed in python packages dir, do you mind to shred some light on that?

carsonwang:
The Spark scripts like spark-submit will always check SPARK_HOME. Previously without this PR, if you have SPARK_HOME set, it will also use configurations under that directory. However it doesn't use the Spark jars under SPARK_HOME correctly. I think you can just test the Spark jar path and make sure it is using the correct one under SPARK_HOME.

carsonwang:
LGTM. Thanks!

PR #253, opened 2022-07-12
update to use Spark 3.3

kira-lin:


kira-lin:
better to use spark shim in 3.3

carsonwang:
We can drop Python 3.6. PySpark 3.3 requires Python 3.7+.

kira-lin:
ok, but we'll need a shim layer to add support for 3.3. I've told @KepingYan to look into it. I 'll drop python 3.6 [here](https://github.com/oap-project/raydp/pull/214), and it seems like raydp can work with ray 1.13.

core/shims/common/src/main/scala/com/intel/raydp/shims/SparkShimLoader.scala
```diff
@@ -0,0 +1,61 @@
+package com.intel.raydp.shims
```
carsonwang, line 1: can you please add apache license header for all new scale files?

core/shims/spark321/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+          <execution>
+            <id>scala-test-compile-first</id>
+            <phase>process-test-resources</phase>
+            <goals>
+              <goal>testCompile</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <!-- <plugin>
```
carsonwang, line 63: I saw commented code in several pom files. Are they still useful? Otherwise clean up them?

core/shims/spark321/src/main/scala/com/intel/raydp/shims/SparkShimProvider.scala
```diff
@@ -0,0 +1,24 @@
+package com.intel.raydp.shims.spark321
+
+import com.intel.raydp.shims.{SparkShims, SparkShimDescriptor}
+
+object SparkShimProvider {
+  val SPARK311_DESCRIPTOR = SparkShimDescriptor(3, 1, 1)
+  val SPARK312_DESCRIPTOR = SparkShimDescriptor(3, 1, 2)
+  val SPARK313_DESCRIPTOR = SparkShimDescriptor(3, 1, 3)
+  val SPARK321_DESCRIPTOR = SparkShimDescriptor(3, 2, 1)
```
carsonwang, line 9: We built this shim with Spark 3.2 but it also works for 3.1?

carsonwang (commented):
Thanks for the work!

core/shims/spark321/src/main/scala/com/intel/raydp/shims/SparkShimProvider.scala
kira-lin: yes, we previously pull some spark class into raydp, and it works for 3.1 and 3.2. So we don't need to implement a 3.1 shim, it would be identical to the 3.2 one

core/shims/common/src/main/scala/com/intel/raydp/shims/SparkShimLoader.scala
kira-lin: ok

core/shims/spark321/pom.xml
kira-lin: ok

carsonwang:
LGTM

PR #255, opened 2022-07-17
Fix Spark download Url in tests

pang-wu:
Previously we are using the download URL for latest Spark release in our tests, which is different than the archived versions. This approach will create problem after a new release come out, because the old release will be removed from dlcdn and added to archive (return 404).


kira-lin:
LGTM, @carsonwang

carsonwang:
Merged. Thanks!

pang-wu:
Thanks all!

PR #258, opened 2022-07-25
Support fractional resource scheduling

pang-wu:
Support fractional CPU and GPU resource scheduling. This PR actually achieve three goals:
* Support schedule multiple executors in one vCPU, with each executor has 1 spark core (for parallism, like how Spark CPU request work on Kubernetes, but no limit), using `spark.ray.actor.resource.cpu` config.
* Support scheduling on GPU nodes, using `spark.ray.actor.resource.gpu` config.
* Support schedule multiple executors in one GPU.

For more details, please refer to [this RFC](https://github.com/oap-project/raydp/issues/259).

### DIFF
--- a/.github/workflows/raydp.yml
+++ b/.github/workflows/raydp.yml
@@ -100,7 +100,13 @@ jobs:
       - name: Test with pytest
         run: |
           ray start --head --num-cpus 4
-          pytest python/raydp/tests/
+          pytest python/raydp/tests/ --deselect python/raydp/tests/test_spark_cluster.py::test_spark_on_fractional_cpu --deselect python/raydp/tests/test_spark_cluster.py::test_spark_on_gpu_machine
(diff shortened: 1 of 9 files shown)


pang-wu:
@carsonwang & team, please kindly let me know if you want a call to discuss this proposal.

carsonwang:
Thanks @pang-wu for the work! How will the gpu config be used as Spark actually is not aware of the gpu resource?

pang-wu:
@carsonwang This is up to the user. So the expectation here is if a cluster has GPU, Spark can still launch executor to executor for non-gpu tasks using this config. Right not deveopers has to use mixed node cluster to do this. User can also put GPU aware code inside spark functions like map or something similar to use GPU resource.. similar to the custom resource scheduling.

carsonwang:
LGTM

Issue #175, opened 2021-09-03
when i create initialization  twice ray,raydp , An error occurred。


chinaboyll:
My version:
Ray Cluster:1.3.0
Raydp:0.3.0
Python: 3.6.4
when i create initialization  twice ray,raydp , An error occurred。
-----------------
Caused by: java.net.BindException: Address already in use (Bind failed)
        at java.net.PlainSocketImpl.socketBind(Native Method)
        at java.net.AbstractPlainSocketImpl.bind(AbstractPlainSocketImpl.java:387)
        at java.net.ServerSocket.bind(ServerSocket.java:375)
        at py4j.GatewayServer.startSocket(GatewayServer.java:786)
-----------------
I set config configs={"spark.ui.port":"xxx"}

carsonwang:
You only need to init once. If you see this error, it is because the pervious driver is not closed properly. You can find the Java process ID and kill it.

Bowen0729:
> You only need to init once. If you see this error, it is because the pervious driver is not closed properly. You can find the Java process ID and kill it.

What if I want to start two spark driver on the same time? What should I do?

carsonwang:
@Bowen0729 , can you please share more about your use case and requirement?

Bowen0729:
Two users both user raydp in the same conda environment, one of them throw exception address already in use

Bowen0729:
> @Bowen0729 , can you please share more about your use case and requirement?

The Java Gateway port is always 25333, why not a random port?

carsonwang:
This sounds like an issue. So two users will run their Spark driver on the same node and connect to the same Ray cluster, right? Are you interested in submitting a PR for this?

Bowen0729:
> This sounds like an issue. So two users will run their Spark driver on the same node and connect to the same Ray cluster, right? Are you interested in submitting a PR for this?

Sure, I will submit a PR recently. Thank you for your reply

carsonwang:
Great! Thanks!

pang-wu:
@Bowen0729 @carsonwang In this case how do we distinguish the user call `init_spark` twice by mistake or by intention?

Bowen0729:
> 

It's a issue about the java gateway always starts with the default port, so when user call init_spark twice in different session,  the new java gateway cannot start. 

when user call init_spark in the same session, raydp will throw "The spark environment has inited".

Hoeze:
[...]

PR #262, opened 2022-07-31
Two users cannot run their Spark driver on the same node and connect to the same Ray cluster.

Bowen0729:
https://github.com/oap-project/raydp/issues/175#issuecomment-1195190398

two users cannot run their Spark driver on the same node and connect to the same Ray cluster. 

Start the java gateway with a random port.

### DIFF
--- a/core/src/main/scala/org/apache/spark/deploy/raydp/AppMasterEntryPoint.scala
+++ b/core/src/main/scala/org/apache/spark/deploy/raydp/AppMasterEntryPoint.scala
@@ -18,6 +18,7 @@
 package org.apache.spark.deploy.raydp
 
 import java.io.{DataOutputStream, File, FileOutputStream}
+import java.net.InetAddress
 import java.nio.file.Files
 
 import py4j.GatewayServer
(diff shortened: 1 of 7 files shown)


carsonwang:
Thank you @Bowen0729 , can you please add a test for this?

Bowen0729:
> Thank you @Bowen0729 , can you please add a test for this?

Hi, I have added a test for this.

Bowen0729:
The CI is failed. What's the problem?

carsonwang:
The CI is flaky these days. I just triggered a rerun of the failed test.

Bowen0729:
I checked the CI log, and I find the Build and install step run mvn verify -q, not mvn clean package -q -DskipTests, so my test is always failed. What should I do?
@carsonwang

Bowen0729:
@carsonwang Hi, I checked the CI log, the Build and install step run mvn verify -q, not the mvn clean package, the raydp jar seems not been compile with the new code, so the CI is always failed, am I right?

Bowen0729:
@carsonwang My test is passed, other tests on macos error.

carsonwang:
@Bowen0729 Previously I saw you tried to test running two spark drivers on the same node, that is the test we want to have. Is there any challenge to do that?

Bowen0729:
> @Bowen0729 Previously I saw you tried to test running two spark drivers on the same node, that is the test we want to have. Is there any challenge to do that?

Yes, the test can passed in my mac, but cannot passed for the CI, I think I need some help.

Bowen0729:
Sorry, I've been  a little busy recently, so I don't have time to solve this problem. Thanks for helping.

kira-lin:
@carsonwang the test still hangs sometime. Shall we skip the tests?

carsonwang:
Ok. Let's skip it for now. Does it also hangs sometime on your local machine?

kira-lin:
yes, it hangs sometime and when I ctrl-c it, it shows popen_fork, even if I should have set start_method to spawn.

carsonwang:
Thanks all! Merged

PR #267, opened 2022-08-26
Upgrade raydp to be compatible with ray 2.0.0

kira-lin:


core/pom.xml
```diff
@@ -20,7 +20,7 @@
     <spark.version>3.2.0</spark.version>
     <fasterxml.jackson.version>2.12.2</fasterxml.jackson.version>
     <scala.binary.version>2.12</scala.binary.version>
-    <ray.version>1.9.0</ray.version>
+    <ray.version>1.13.0</ray.version>
```
carsonwang, line 23: Does this have to upgraded? Is it possible to support from 1.9.0 to 2.0.0?

kira-lin: 2.0.0 jar is not released yet. Do you mean we just use 1.9.0 to compile?

carsonwang: yes, will it work?

pang-wu:
We might also want to migrate off the usage of `ray.worker.global_worker.node`, they are [marked as deprecated](https://github.com/ray-project/ray/blob/ray-2.0.0/python/ray/__init__.py#L201). But right now I didn't find an alternative..

kira-lin:
@pang-wu Yes, I considered to replace it. But the key point is we have to use _deserialize_and_register in `global_worker.core_worker` now.

pang-wu:
> @pang-wu Yes, I considered to replace it. But the key point is we have to use _deserialize_and_register in `global_worker.core_worker` now.

@kira-lin Isn't that still using `ray.worker.*`? I think we might need to migrate off from using the whole `ray._private.*` and rely on public API?

kira-lin:
Yes, we'll need to make some upstream changes. For now we can't only use public APIs, we would not be able to call from_spark in that case.

pang-wu:
gotcha, thanks for the prompt responses!

carsonwang:
Thanks @kira-lin ! Merging this.

PR #269, opened 2022-09-07
Fix the pyspark driver log level

carsonwang:
Because Ray's jar is added in "spark.driver.extraClassPath", spark driver's SL4j binds the jar from Ray's jar instead of the one in Spark. This causes many INFO logs on the driver. It also makes `sparkContext.setLogLevel` stop working. As a workaround, we can add the spark slf4j-log4j on the driver extraClassPath first so the driver can still binds to the correct jar.

There are still Spark app master INFO logs printed on the driver as Ray redirects the actor logs to driver. To disable it, we can set the configuration in Ray init i.e. `ray.init(log_to_driver=False)`.

### DIFF
--- a/doc/spark_on_ray.md
+++ b/doc/spark_on_ray.md
@@ -60,3 +60,6 @@ spark = raydp.init_spark(...,enable_hive=True)
 spark.sql("select * from db.xxx").show()
 ```
 
+### Logging
++ Driver Log: By default, the spark driver log level is WARN. After getting a Spark session by running `spark = raydp.init_spark`, you can change the log level for example `spark.sparkContext.setLogLevel("INFO")`. You will also see some AppMaster INFO logs on the driver. This is because Ray redirects the actor logs to driver by default. To disable logging to driver, you can set it in Ray init `ray.init(log_to_driver=False)`
++ Executor Log: The spark executor logs are stored in Ray's logging directory. By default they are available at /tmp/ray/session_\*/logs/java-worker-\*.log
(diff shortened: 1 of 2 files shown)


carsonwang:
@kira-lin , please help take a look.

kira-lin:
LGTM. Shall we add some doc to explain? Disabling all logs to driver might be an overkill

carsonwang:
Right, we'll let the user to set the `log_to_driver` configuration or not. Let me add the document.

doc/spark_on_ray.md
```diff
@@ -60,3 +60,6 @@ spark = raydp.init_spark(...,enable_hive=True)
 spark.sql("select * from db.xxx").show()
 ```
 
+### Logging
++ Driver Log: By default, the spark driver log level is WARN. After getting a Spark session by running `spark = raydp.init_spark`, you can change the log level for example `spark.sparkContext.setLogLevel("INFO")`. You will also see some AppMaster INFO logs on the driver. This is because Ray redirects the actor logs to driver by default. To disable logging to driver, you can set it in Ray init `ray.init(log_to_driver=False)`
```
pang-wu, line 64: I have some second thoughts on suggesting user to `log_to_driver` though, one of the biggest benefit on using raydp is I can print in spark functions and see the logs in driver, but setting `log_to_driver` will disable that all together. Is there a way to just disable to app master info log so we get a better notebook experience? @carsonwang

carsonwang: Thanks for the suggestion, @pang-wu . We will take a look how to support it.

PR #278, opened 2022-10-19
Print warning message when spark.dynamicAllocation.maxExecutors is not set

kira-lin:
Print a warning message if dynamic allocation is enabled but maxExecutors is not set. The reason for this is because spark might request too many executors and crash ray if it is not set.


python/raydp/spark/ray_cluster.py
```diff
@@ -88,6 +88,14 @@ def get_spark_session(self,
         else:
             extra_conf[driver_cp_key] = driver_cp
         spark_builder = SparkSession.builder
+        dyn_alloc_key = "spark.dynamicAllocation.enabled"
+        if dyn_alloc_key in extra_conf and extra_conf[dyn_alloc_key] == "true":
+            max_executor_key = "spark.dynamicAllocation.maxExecutors"
+            # set max executors if not set. otherwise spark might request too many actors
+            if max_executor_key not in extra_conf:
+                total_cores = ray.cluster_resources()["CPU"]
```
carsonwang, line 96: It seems this doesn't work with Ray's auto scaling. It is common that people just start a single head node at the beginning in cloud.

kira-lin: Oh, right. But still, if they want to use auto scaling, I think they should set a max value, otherwise spark can request as many executors as wanted, and user may cost too much money? In case of auto scaling, how can we know when to stop spark to request more executors?

carsonwang: Right, we still need users to configure it. What about just print a warning if the configuration is missing?

python/raydp/spark/ray_cluster.py
```diff
@@ -88,6 +88,12 @@ def get_spark_session(self,
         else:
             extra_conf[driver_cp_key] = driver_cp
         spark_builder = SparkSession.builder
+        dyn_alloc_key = "spark.dynamicAllocation.enabled"
+        if dyn_alloc_key in extra_conf and extra_conf[dyn_alloc_key] == "true":
+            max_executor_key = "spark.dynamicAllocation.maxExecutors"
+            # set max executors if not set. otherwise spark might request too many actors
+            if max_executor_key not in extra_conf:
+                print("Warn: Consider setting max executor number to match the cluster config.")
```
carsonwang, line 96: Better to be more specific here. Let's directly call out setting spark.dynamicAllocation.maxExecutors in the message?

carsonwang:
Can you please also update the tile and description of the PR?

carsonwang:
LGTM

PR #283, opened 2022-10-26
Add more sparkversions to shim layer

kira-lin:
3.2.0, 3.2.2 and 3.3.1 is added. From now on, we need to add spark version to the shim provider once it releases a new one, otherwise it cannot match a provider.


carsonwang:
LGTM

PR #286, opened 2022-11-04
port to new ray train api in estimators

kira-lin:
Currently used ray.train.Trainer has been deprecated. In addition, the get_model method is not working now. This PR updates to use TorchTrainer and TensorflowTrainer and implements get_model based on checkpoint.


python/raydp/torch/estimator.py
```diff
@@ -211,15 +215,15 @@ def train_func(config):
         metrics = config["metrics"]
 
         # create dataset
-        train_data_shard = get_dataset_shard("train")
+        train_data_shard = session.get_dataset_shard("train")
```
carsonwang, line 218: The latest Ray document suggests using `iter_torch_batches` instead of `to_torch` on the data shard. Can you please take a look at the difference?

carsonwang:
@KepingYan , can you please take a look as well?

KepingYan (commented):
LGTM.

python/raydp/torch/estimator.py
```diff
@@ -334,17 +353,14 @@ def fit_on_spark(self,
...
-
-    def restore(self, checkpoint):
-        assert self._trainer is not None, "Must call fit first"
-        self._trainer.load(checkpoint)
-
-    def shutdown(self):
-        if self._trainer is not None:
-            self._trainer.shutdown()
-            self._trainer = None
+        states = self._trained_results.checkpoint.to_dict()["state_dict"]
```
carsonwang, line 356: Why not calling `checkpoint.get_model()` to get the model?

python/raydp/torch/estimator.py
```diff
@@ -233,12 +235,20 @@ def train_func(config):
...
                                                                             model, loss, metrics)
-                train.report(epoch=epoch, evaluate_res=evaluate_res, test_loss=evaluate_loss)
+                session.report(dict(epoch=epoch, eval_res=eval_res, test_loss=evaluate_loss))
                 loss_results.append(evaluate_loss)
+        if hasattr(model, "module"):
+            states = model.module.state_dict()
+        else:
+            # if num_workers = 1, model is not wrapped
+            states = model.state_dict()
+        session.report({}, checkpoint=Checkpoint.from_dict({
```
carsonwang, line 249: This saves the checkpoint for every epoch. Is this a default behavior? Otherwise we also don't need to do it by default.

python/raydp/torch/estimator.py
kira-lin: To use get_model, we need to use framework specific checkpoint, namely TorchCheckpoint and TensorflowCheckpoint, these are in "public api beta" phase. They do the same thing as we do here.

python/raydp/torch/estimator.py
kira-lin: I tested removing this, but the checkpoint included in the returned result does not contain anything.

python/raydp/tf/estimator.py
```diff
@@ -45,13 +49,16 @@ def __init__(self,
                  label_shape: Optional[tf.TensorShape] = None,
                  batch_size: int = 128,
                  num_epochs: int = 1,
-                 shuffle: bool = True,
+                 shuffle: bool = False,
```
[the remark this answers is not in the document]
kira-lin: Do we want to make this default? this might hurt performance for users who don't need it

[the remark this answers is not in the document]
carsonwang: Yes, let's keep the default as "True". It is common to do a global shuffle before training. Other frameworks like horovod estimator set the default to true as well.

python/raydp/torch/estimator.py
carsonwang: Where is checkpoint stored? Are anything in the path?

python/raydp/torch/estimator.py
carsonwang: I see. Let's use our own code for now.

python/raydp/tf/estimator.py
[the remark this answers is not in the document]
kira-lin: ok

python/raydp/torch/estimator.py
kira-lin: It is stored in ~/ray_results, there are some pickled binary files. But when I use TorchCheckpoint.from_directory to load it, get_model does not work.

PR #287, opened 2022-11-18
Implement resource based node affinity for Spark Driver

pang-wu:
### What Problem Does this PR Solve
By default the `SPARK_MASTER` and `OBJ_HOLDER` actor could be placed at any node -- This is not a problem when you have a large head node, or all nodes are on demand in the cluster. However in our production setup, we are using small head node with spot machines as workers. Spark master will get launch on worker nodes and when we get spot kill the driver died and the job hang.

To solve this problem, we need a way to do node affinity: We need a way to force the Spark driver (or any other critical Spark actor infrastructure) to land on stable machines that is defined by users.

### Purposed Solution

Add a extra config in raydp python API for both Spark master so user can place them to nodes with resources specified:
```
spark.ray.raydp_spark_master.resource.*
```

Example usage:

```python
raydp.init_spark(...,
configs = {
    # ... other configs
    'spark.ray.raydp_spark_master.resource.CPU': 0,
    'spark.ray.raydp_spark_master.resource.spark_master': 1,  # Force Spark driver related actor run on headnode
})
```

In cluster config yaml:
```yaml
available_node_types:
  ray.head.default:
    resources:
      CPU: 0    # We intentionally set this to 0 so not executors are on headnode
      spark_master: 100  # Just gave it a large enough number so all drivers are there
```

### DIFF
--- a/doc/spark_on_ray.md
+++ b/doc/spark_on_ray.md
@@ -1,3 +1,26 @@
+### How to schedule Spark Master to a designated node
+
+RayDP will create a ray actor called `RayDPSparkMaster`, which will then launch the java process, acting like a Master in a tradtional Spark cluster. By default, this actor could be scheduled to any node in the ray cluster. If you want it to be on a particular node, you can assign some custom resources to that node, and request those resources when starting `RayDPSparkMaster` by setting `spark.ray.raydp_spark_master.resource.*` in `init_spark`.
+
+As an example:
+
+```python
+raydp.init_spark(...,
+configs = {
+    # ... other configs
+    'spark.ray.raydp_spark_master.resource.CPU': 0,
+    'spark.ray.raydp_spark_master.resource.spark_master': 1,  # Force Spark driver related actor run on headnode
+})
+```
+
+In cluster config yaml:
+```yaml
+available_node_types:
+  ray.head.default:
+    resources:
+      CPU: 0    # We intentionally set this to 0 so not executors are on headnode
+      spark_master: 100  # Just gave it a large enough number so all drivers are there
+```
 
 ### External Shuffle Service & Dynamic Resource Allocation
 
--- a/doc/spark_on_ray.md
+++ b/doc/spark_on_ray.md
@@ -1,4 +1,4 @@
-### How to schedule Spark Master to a designated node
+### Spark master node affinity
 
 RayDP will create a ray actor called `RayDPSparkMaster`, which will then launch the java process, acting like a Master in a tradtional Spark cluster. By default, this actor could be scheduled to any node in the ray cluster. If you want it to be on a particular node, you can assign some custom resources to that node, and request those resources when starting `RayDPSparkMaster` by setting `spark.ray.raydp_spark_master.resource.*` in `init_spark`.
 
@@ -18,7 +18,7 @@ In cluster config yaml:
 available_node_types:
   ray.head.default:
     resources:
-      CPU: 0    # We intentionally set this to 0 so not executors are on headnode
+      CPU: 0    # We intentionally set this to 0 so no executor is on headnode
       spark_master: 100  # Just gave it a large enough number so all drivers are there
 ```
 
--- a/python/raydp/spark/ray_cluster.py
+++ b/python/raydp/spark/ray_cluster.py
@@ -44,9 +44,9 @@ class SparkCluster(Cluster):
 
         if resources:
             num_cpu = 1
-            if 'CPU' in resources:
-                num_cpu = resources['CPU']
-                resources.pop('CPU', None)
+            if "CPU" in resources:
+                num_cpu = resources["CPU"]
+                resources.pop("CPU", None)
             self._spark_master_handle = RayDPSparkMaster.options(name=spark_master_name,
                                                                  num_cpus=num_cpu,
                                                                  resources=resources) \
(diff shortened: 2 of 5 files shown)


carsonwang:
Thanks @pang-wu for the PR. Could you please submit the PR to master to trigger the CI? The object holder is an internal concept. Do you think we can use a single config for both spark master and object holder so the user don't need to know anything about the object holder? cc @kira-lin

pang-wu:
@carsonwang will do!

pang-wu:
Actually I have some second thoughts on using the same config, I think that will be confusing for the user because they don't know the object holder exists -- they might schedule 1 CPU for spark master but it ended up taking 2 and the job cannot start if say the head node doesn't have sufficient resource, what do you think @carsonwang ?

kira-lin:
Yes. How about setting num_cpus to 0 for obj_holder? It doesn't perform much calculation anyway.
Also, as stated in ray doc

> By default, Ray tasks use 1 CPU resource and Ray actors use 1 CPU for scheduling and 0 CPU for running (This means, by default, actors cannot get scheduled on a zero-cpu node, but an infinite number of them can run on any non-zero cpu node. 

So using the same config might just work. The problem is that users need to pay attention not to take up all custom resources for the master, e.g. `spark_master` resource in your description

pang-wu:
@kira-lin I think the problem here is the total amount of resource for master actors is implicit because there are two actors under the hood -- setting up CPU to 0 won't solve the problem if the user use other resource for scheduling (like the `spark_master` in my example), the total amount defined in the node needs to be 2x of what get specified in the config -- this is a piece of knowledge the user won't know if he didn't read our code and will be a surprise for the user

kira-lin:
Right. How about merging obj_holder into spark master? I think we can make them the same actor. cc @carsonwang

pang-wu:
> Right. How about merging obj_holder into spark master? I think we can make them the same actor. cc @carsonwang

Yeah, I think that make sense if all of them are supposed to be in the driver.

pang-wu:
@kira-lin @carsonwang merged from master and consolidate to one config, would you please take a look again?

python/raydp/spark/ray_cluster.py
```diff
@@ -35,16 +35,38 @@ def __init__(self, app_name, configs):
...
+                .remote(self._configs)
+
         ray.get(self._spark_master_handle.start_up.remote())
 
+    def _get_master_resources(self, configs: Dict[str, str]) -> Dict[str, float]:
+        resources = {}
+        object_holder_config_prefix = "spark.ray.raydp_spark_master.resource."
+        for key in configs:
+            if key.startswith(object_holder_config_prefix):
+                resource_name = key[len(object_holder_config_prefix):]
```
kira-lin, line 65:
```suggestion
        spark_master_config_prefix = "spark.ray.raydp_spark_master.resource."
        for key in configs:
            if key.startswith(spark_master_config_prefix):
                resource_name = key[len(spark_master_config_prefix):]
```

kira-lin (commented):
Please rename the configs, others LGTM

pang-wu:
done

kira-lin:
I added some doc @carsonwang pls take a look

doc/spark_on_ray.md
```diff
@@ -1,3 +1,26 @@
...
+    'spark.ray.raydp_spark_master.resource.spark_master': 1,  # Force Spark driver related actor run on headnode
+})
+```
+
+In cluster config yaml:
+```yaml
+available_node_types:
+  ray.head.default:
+    resources:
+      CPU: 0    # We intentionally set this to 0 so not executors are on headnode
```
pang-wu, line 21: There is a typo here `no executor is` instead of `not executors are`?

doc/spark_on_ray.md
```diff
@@ -1,3 +1,26 @@
+### How to schedule Spark Master to a designated node
```
pang-wu, line 1: What do you think about `Spark master node affinity`? I think this allows you to send drivers to a set of nodes, not only one.

pang-wu: In fact, I am thinking maybe we should do this for executors as well, do different job's executor can land on different worker nodes. This is a unique feature for Spark on Ray and it is useful for people to use one cluster host multiple jobs with different resource requirement.

kira-lin: > What do you think about Spark master node affinity? I think this allows you to send drivers to a set of nodes, not only one.

Hmm, I was thinking as a normal user, who just cares about one application at a time. But yes, your suggested name is also fine.

Doing this for executors is trickier, because each executor can require different resource? Or we limit it to same resource for every executor.

pang-wu: > Hmm, I was thinking as a normal user, who just cares about one application at a time.

Yeah in that case the user might not even need this PR because most likely his/her driver node will be run on headnode. But when the Ray cluster is crowded, it is a different story :)

> Doing this for executors is trickier, because each executor can require different resource? Or we limit it to same resource for every executor.

So in our usecase we have all the different jobs belong to one batch automation pipeline usecase all run in one Ray cluster, in this case we save the cluster launch/tear down time. For executor, what we need is give use a custom config to assign extra custom resources, so the user can assign the executors for one particular job to a node config (and one Ray cluster might have multiple node configs for different types of wirker nodes).

kira-lin: Sure, let's do it in another pr.

PR #288, opened 2022-11-23
Merge the function of Object Holder into RayDPSparkMaster

kira-lin:
Object holder is added when RayDPSparkMaster is not an actor. Now that we already have an actor at driver node, we can merge the object holding function into it.

### DIFF
--- a/python/raydp/context.py
+++ b/python/raydp/context.py
@@ -119,13 +119,13 @@ class _SparkContext(ContextDecorator):
             self._configs)
         return self._spark_session
 
-    def stop(self, del_obj_holder=True):
+    def stop(self, cleanup_data=True):
         if self._spark_session is not None:
             self._spark_session.stop()
             self._spark_session = None
         if self._spark_cluster is not None:
-            self._spark_cluster.stop(del_obj_holder)
-            if del_obj_holder:
+            self._spark_cluster.stop(cleanup_data)
+            if cleanup_data:
                 self._spark_cluster = None
         if self._placement_group_strategy is not None:
             if self._placement_group is not None:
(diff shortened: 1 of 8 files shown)


kira-lin:
Please review @carsonwang @pang-wu

pang-wu:
👍

python/raydp/context.py
```diff
@@ -126,12 +123,10 @@ def stop(self, del_obj_holder=True):
         if self._spark_session is not None:
             self._spark_session.stop()
             self._spark_session = None
-        if self.handle is not None and del_obj_holder is True:
-            self.handle.terminate.remote()
-            self.handle = None
         if self._spark_cluster is not None:
-            self._spark_cluster.stop()
-            self._spark_cluster = None
+            self._spark_cluster.stop(del_obj_holder)
```
carsonwang, line 127: The API `raydp.stop_spark(del_obj_holder=True)` is user facing. Any idea for a better name to replace name `del_obj_holder`?

kira-lin: Maybe `raydp.stop_spark(cleanup_data=True)`?

pang-wu: I am thinking shall we give it a default value?

kira-lin: Hmm, giving a default value keeps the behavior same as before. Otherwise users might upgrade raydp and find out the scripts are failing.

pang-wu: right, but I think we don't have a default right now, or am i missing something?

kira-lin: Yes, we have a default value now, we're talking if we should change the option name

carsonwang: It sounds better. Please also add some document to avoid confusion.

carsonwang:
LGTM

Issue #299, opened 2023-01-24
Initial job has not accepted any resources...


Tellz777:
Hi, guys. 

I have Ray-cluster installed in Kubernetes:

```
1 master (2 CPU/10GB memory)
4 workers (2 CPU/10GB memory each)
```
It uses a custom image of Ray I've builded. 

Dockerfile:

```
FROM rayproject/ray:2.2.0
USER root
RUN apt-get update && apt-get install -y openjdk-11-jdk \
    && rm -rf /var/lib/apt/lists/*
ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
RUN pip install raydp==1.5.0
```
Im trying to run your example of spark job on my remote Ray cluster (locally it works perfect).

```
import ray
import raydp

ray.init(address='ray://<my-remote-kuberay-head-svc>:10001')

spark = raydp.init_spark(app_name='RayDP Example', num_executors=1, executor_cores=1, executor_memory='1GB')
spark.sparkContext.setLogLevel("INFO")
df = spark.createDataFrame([('look',), ('spark',), ('tutorial',), ('spark',), ('look', ), ('python', )], ['word'])
df.show()
word_count = df.groupBy('word').count()
word_count.show()

raydp.stop_spark()
```

And starting to receive this WARN message repeatedly:  "Initial job has not accepted any resources; check your cluster UI to ensure that workers are registered and have sufficient resources"

As I see from the logs it starts Spark UI locally on my laptop here: http://127.0.0.1:4040/

Spark master logs:
```
:actor_name:RayDPSparkMaster
Setting default log level to "WARN".
To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/ray/anaconda3/lib/python3.7/site-packages/pyspark/jars/log4j-slf4j-impl-2.17.2.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/ray/anaconda3/lib/python3.7/site-packages/ray/jars/ray_dist.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.apache.logging.slf4j.Log4jLoggerFactory]
```
Ray app master logs:
```
[...]

PR #301, opened 2023-02-01
Set all spark conf when creating RayAppMaster

kira-lin:
This PR tries to solve #299 
Signed-off-by: Zhi Lin <zhi.lin@intel.com>


carsonwang:
Can you please add a test?

kira-lin:
we can't actually test the case in #299 because ci has only one node, and we've been testing ray client for every test. We can try to print spark conf to check spark.driver.host though.

pang-wu:
Will this fix the bug when setting `spark.jars.packages` the jars won't be download in worker nodes?

kira-lin:
> Will this fix the bug when setting spark.jars.packages the jars won't be download in worker nodes?

I think it won't, because this error occurs before the fault tolerance PR, but you can test it. Sorry we don't have enough bandwidth to look into it, I think we have a work-around for this, right?

pang-wu:
Yeah correct, no worries.

PR #309, opened 2023-02-24
When converting spark df to ray dataset, you can choose whether to release the resources requested by spark

hezhaozhao-git:
**Restore the resources occupied by the park, so that the subsequent tasks can use the CPU, memory and other resources used by the previous spark**

When init spark uses a lot of all available resources, if you call ray.data.from_spark(df), the task will be blocked at this time, and ray will give us a warning log to increase resources, as shown in the following figure:
企业微信截图_16772426688561

Add a parameter recover_ray_resources_from_spark in ray.data.from_spark. The default value is False, which means that resources are not restored. In this way, users can choose whether to stop sparkSession during the process of converting spark df to ray dataset, so as to release the resources occupied by spark , can convert spark dataframe into ray dataset smoothly, but this will bring another problem, that is, spark and existing dataframe and RDD objects will be unavailable. At this time, you need to call raydp.stop_spark() to stop spark. You also need to use spark, you can re-init_spark

### DIFF
--- a/python/raydp/spark/dataset.py
+++ b/python/raydp/spark/dataset.py
@@ -161,7 +161,7 @@ def _save_spark_df_to_object_store(df: sql.DataFrame, use_batch: bool = True,
     record_tuples = [(record.objectId(), record.ownerAddress(), record.numRecords())
                         for record in records]
     blocks, block_sizes = _register_objects(record_tuples)
-    if recover_ray_resource_from_spark:
+    if recover_ray_resources_from_spark:
         df.sql_ctx.sparkSession.stop()
     if _use_owner is True:
         holder = ray.get_actor(obj_holder_name)
--- a/python/raydp/spark/dataset.py
+++ b/python/raydp/spark/dataset.py
@@ -173,13 +173,13 @@ def _save_spark_df_to_object_store(df: sql.DataFrame, use_batch: bool = True,
 
 def spark_dataframe_to_ray_dataset(df: sql.DataFrame,
                                    parallelism: Optional[int] = None,
-                                   _use_owner: bool = False, 
+                                   _use_owner: bool = False,
                                    recover_ray_resources_from_spark=False):
     num_part = df.rdd.getNumPartitions()
     if parallelism is not None:
         if parallelism != num_part:
             df = df.repartition(parallelism)
-    blocks, _ = _save_spark_df_to_object_store(df, False, _use_owner, 
+    blocks, _ = _save_spark_df_to_object_store(df, False, _use_owner,
                                                recover_ray_resources_from_spark)
     return from_arrow_refs(blocks)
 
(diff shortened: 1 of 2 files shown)


hezhaozhao-git:
@kira-lin This is not a bug in the code I submitted, what should I do? My test case has passed.

kira-lin:
You can call `spark_dataset_to_ray_dataset`, and then `raydp.stop_spark(cleanup_data=False)`. Is this not enough for you? Your code will leave our code in an inconsistent state, i.e. SparkCluster still think its spark session is alive.

hezhaozhao-git:
> raydp.stop_spark(cleanup_data=False)

kira-lin:
After calling spark_dataset_to_ray_dataset, you can close the spark session by raydp.stop_spark(cleanup_data=False). You also need to set _use_owner when calling spark_dataset_to_ray_dataset.
I see, your problem is because ray.data.from_arrow_refs need some CPU to be scheduled, but all CPUs have been taken up by spark. What version of ray are you using? I think ray actor only require CPU when scheduling, after it starts it should release the resources. As a work-around, you can claim you have more CPU than you actually have when starting the ray cluster if possible.

hezhaozhao-git:
> 调用spark_dataset_to_ray_dataset后，可以通过raydp.stop_spark(cleanup_data=False)关闭spark session。调用spark_dataset_to_ray_dataset时也需要设置_use_owner。 我明白了，你的问题是因为 ray.data.from_arrow_refs 需要一些 CPU 来调度，但是所有的 CPU 都被 spark 占用了。您使用的是什么版本的射线？我认为ray actor在调度时只需要CPU，启动后应该释放资源。作为变通方法，如果可能的话，您可以在启动射线簇时声称您拥有比实际拥有的更多的 CPU。

The ray actor does occupy resources when it is in use, but if spark has applied for all the resources, calling ray.data.from_spark at this time will not convert the spark df into a ray dataset. Although calling stop_spark can indeed release resources, but After the resource is released, the ray.data.from_spark call reports an error. I read the source code and found that the data conversion process uses ray put and ray get. After the spark df object is put, the sparkSession may no longer be needed at this time. At this time, the resources occupied by spark can be released, and then the ray actor There are schedulable resources, so I have the previous PR

hezhaozhao-git:
> After calling spark_dataset_to_ray_dataset, you can close the spark session by raydp.stop_spark(cleanup_data=False). You also need to set _use_owner when calling spark_dataset_to_ray_dataset. I see, your problem is because ray.data.from_arrow_refs need some CPU to be scheduled, but all CPUs have been taken up by spark. What version of ray are you using? I think ray actor only require CPU when scheduling, after it starts it should release the resources. As a work-around, you can claim you have more CPU than you actually have when starting the ray cluster if possible.

The ray actor does occupy resources when it is in use, but if spark has applied for all the resources, calling ray.data.from_spark at this time will not convert the spark df into a ray dataset. Although calling stop_spark can indeed release resources, but After the resource is released, the ray.data.from_spark call reports an error. I read the source code and found that the data conversion process uses ray put and ray get. After the spark df object is put, the sparkSession may no longer be needed at this time. At this time, the resources occupied by spark can be released, and then the ray actor There are schedulable resources, so I have the previous PR

PR #312, opened 2023-03-02
Incorrect Jar path in setup.py

zjf2012:
limit direct folder of jars to "target" so that jars from other folders are not included when glob


zjf2012:
@carsonwang please help review.

carsonwang:
LGTM. Thanks for the fix.

PR #326, opened 2023-04-01
Pick up log4j config files from environment variable if not set in config override

pang-wu:
Enable system admin or infra team to set log4j config file from `SPARK_LOG4J_CONFIG_FILE_NAME` and `RAY_LOG4J_CONFIG_FILE_NAME` environment variable. So the user don't have to know these system details.

### DIFF
--- a/python/raydp/spark/ray_cluster.py
+++ b/python/raydp/spark/ray_cluster.py
@@ -131,11 +131,22 @@ class SparkCluster(Cluster):
         self._configs[SPARK_RAY_LOG4J_FACTORY_CLASS_KEY] = versions.RAY_LOG4J_VERSION
 
         if SPARK_LOG4J_CONFIG_FILE_NAME not in self._configs:
+            # If the config is not passed in by the user, check
+            # by environment variable SPARK_LOG4J_CONFIG_FILE_NAME. This
+            # will give the system admin/infra team a chance to set the
+            # value for all users.
             self._configs[SPARK_LOG4J_CONFIG_FILE_NAME] =\
-                versions.SPARK_LOG4J_CONFIG_FILE_NAME_DEFAULT
+                os.environ.get("SPARK_LOG4J_CONFIG_FILE_NAME",
+                               versions.SPARK_LOG4J_CONFIG_FILE_NAME_DEFAULT)
 
         if RAY_LOG4J_CONFIG_FILE_NAME not in self._configs:
-            self._configs[RAY_LOG4J_CONFIG_FILE_NAME] = versions.RAY_LOG4J_CONFIG_FILE_NAME_DEFAULT
+            # If the config is not passed in by the user, check
+            # by environment variable RAY_LOG4J_CONFIG_FILE_NAME. This
+            # will give the system admin/infra team a chance to set the
+            # value for all users.
+            self._configs[RAY_LOG4J_CONFIG_FILE_NAME] =\
+                os.environ.get("RAY_LOG4J_CONFIG_FILE_NAME",
+                               versions.RAY_LOG4J_CONFIG_FILE_NAME_DEFAULT)
 
         prefer_cp = []
         if SPARK_PREFER_CLASSPATH in self._configs:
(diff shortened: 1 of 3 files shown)


python/raydp/spark/ray_cluster.py
```diff
@@ -131,11 +131,22 @@ def _prepare_spark_configs(self):
         self._configs[SPARK_RAY_LOG4J_FACTORY_CLASS_KEY] = versions.RAY_LOG4J_VERSION
 
         if SPARK_LOG4J_CONFIG_FILE_NAME not in self._configs:
+            # If the config is not passed in by the user, check
```
zjf2012, line 134: It's good idea to get config from env.

python/raydp/spark/ray_cluster.py
```diff
@@ -131,11 +131,22 @@ def _prepare_spark_configs(self):
...
+            # by environment variable SPARK_LOG4J_CONFIG_FILE_NAME. This
+            # will give the system admin/infra team a chance to set the
+            # value for all users.
             self._configs[SPARK_LOG4J_CONFIG_FILE_NAME] =\
-                versions.SPARK_LOG4J_CONFIG_FILE_NAME_DEFAULT
+                os.environ.get("SPARK_LOG4J_CONFIG_FILE_NAME",
```
zjf2012, line 139: Can you remove double quotes from "SPARK_LOG4J_CONFIG_FILE_NAME" ? SPARK_LOG4J_CONFIG_FILE_NAME evaluates to "spark.log4j.config.file.name" which is used in configs in init_spark function. It'll make config name consistent.

python/raydp/spark/ray_cluster.py
```diff
@@ -131,11 +131,22 @@ def _prepare_spark_configs(self):
...
+                               versions.SPARK_LOG4J_CONFIG_FILE_NAME_DEFAULT)
 
         if RAY_LOG4J_CONFIG_FILE_NAME not in self._configs:
-            self._configs[RAY_LOG4J_CONFIG_FILE_NAME] = versions.RAY_LOG4J_CONFIG_FILE_NAME_DEFAULT
+            # If the config is not passed in by the user, check
+            # by environment variable RAY_LOG4J_CONFIG_FILE_NAME. This
+            # will give the system admin/infra team a chance to set the
+            # value for all users.
+            self._configs[RAY_LOG4J_CONFIG_FILE_NAME] =\
+                os.environ.get("RAY_LOG4J_CONFIG_FILE_NAME",
```
zjf2012, line 148: Same as above, RAY_LOG4J_CONFIG_FILE_NAME evaluates to spark.ray.log4j.config.file.name.

zjf2012:
@pang-wu Could you please also update the doc accordingly in doc/ray_on_spark.md? You can scroll down and update the last item. Or add a new item for logging. thanks

python/raydp/spark/ray_cluster.py
pang-wu: What I am trying to do here is if there is an environment variable `SPARK_LOG4J_CONFIG_FILE_NAME` defined, set the value of `spark.log4j.config.file.name` to the value defined by that env var. So I think the logic here is correct.

python/raydp/spark/ray_cluster.py
pang-wu: Same as above. This is just to enable the lib pull in variables set externally.

python/raydp/spark/ray_cluster.py
zjf2012: Yes, I know.  
So you want to use different names for the same thing? SPARK_LOG4J_CONFIG_FILE_NAME for env and "spark.log4j.config.file.name" for configs in init_spark. And they both are for the same log4j file name.
Is it better to use same name, like "spark.log4j.config.file.name"?

pang-wu: I don't think we can use dot(.) in environment variable name, at least [not for bash](https://unix.stackexchange.com/questions/93532/exporting-a-variable-with-dot-in-it). Also using lowercase env name in general is not a convention.

I updated the readme, you can see the example usage.

pang-wu:
> @pang-wu Could you please also update the doc accordingly in doc/ray_on_spark.md? You can scroll down and update the last item. Or add a new item for logging. thanks

done. I also updated `raydp-submit`

python/raydp/spark/ray_cluster.py
zjf2012: ok, I see.

zjf2012:
LGTM

carsonwang:
Merged. Thanks!

PR #328, opened 2023-04-04
[raydp-327] rename ray JVM log file to avoid being monitored and polled

zjf2012:
As stated in https://github.com/ray-project/ray/pull/33797, it's not necessary for ray monitoring ray JVM logs.

This ticket is to make corresponding changes in raydp by setting JVM log file prefix to 'raydp-java-worker'.
The ray PR will be merged and released later. But this ticket can work with and without the PR.

Note: PR https://github.com/ray-project/ray/pull/33797 and this PR don't change any existing raydp behavior in terms of logs printed in spark-shell console. 

The JVM log file here refers to programmatically generated spark executor log file which is different than STDOUT and STDERR point to raylet.out and raylet.err respectively. When you log something, like 'log.info("Finished task....")', in executor, it prints to the JVM log file. But when you log without log4j , like 'print("xxxx")', it prints to STDOUT, that is raylet.out which will be in turn sent to user's console.

### DIFF
--- a/core/agent/src/main/java/org/apache/spark/raydp/Agent.java
+++ b/core/agent/src/main/java/org/apache/spark/raydp/Agent.java
@@ -69,6 +69,9 @@ public class Agent {
       System.setErr(DEFAULT_ERR_PS);
       System.setOut(DEFAULT_OUT_PS);
     }
+    // below is to write ':job_id:<jobid>' to first line of log file prefixed with 'java-worker' as required by
+    // PR, https://github.com/ray-project/ray/pull/31772.
+    // It's a workaround of the ray 2.3.[0-1] issue going to be fixed by https://github.com/ray-project/ray/pull/33665.
     String jobId = System.getenv("RAY_JOB_ID");
     String rayAddress = System.getProperty("ray.address");
     if (jobId != null && rayAddress != null) {
--- a/core/agent/src/main/java/org/apache/spark/raydp/Agent.java
+++ b/core/agent/src/main/java/org/apache/spark/raydp/Agent.java
@@ -71,8 +71,10 @@ public class Agent {
     }
     String jobId = System.getenv("RAY_JOB_ID");
     String rayAddress = System.getProperty("ray.address");
+    String prefix = System.getProperty("ray.logging.file-prefix", "java-worker");
     if (jobId != null && rayAddress != null) {
-      try (FileWriter writer = new FileWriter(logDir + "/java-worker-" + jobId + "-" + pid + ".log")) {
+      try (FileWriter writer = new FileWriter(logDir + "/" + prefix + "-" +
+          jobId + "-" + pid + ".log")) {
         writer.write(":job_id:" + jobId + "\n");
       }
     }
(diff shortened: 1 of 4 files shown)


zjf2012:
@kira-lin please help review.

zjf2012:
@carsonwang just updated the description. Please check.

kira-lin:
Do we need to tell users about this? @carsonwang Maybe add a line to readme, because this is not aligned with Ray, and tell users how to find the executor logs. 

Others LGTM

zjf2012:
> Do we need to tell users about this? @carsonwang Maybe add a line to readme, because this is not aligned with Ray, and tell users how to find the executor logs.
> 
> Others LGTM

As an alternative, we can mention it in our release note since not everyone cares about the executor log.

core/raydp-main/src/main/java/org/apache/spark/raydp/SparkOnRayConfigs.java
```diff
@@ -88,4 +88,7 @@ public class SparkOnRayConfigs {
      * only.
      */
     public static final String RAY_PREFER_CLASSPATH = "spark.ray.preferClassPath";
+
+    public static final String RAYDP_LOGFILE_PREFIX_CFG =
```
carsonwang, line 92: can you please add comment why we set this file prefix?

core/agent/src/main/java/org/apache/spark/raydp/Agent.java
```diff
@@ -72,8 +72,12 @@ public static void premain(String agentArgs, Instrumentation inst)
     String jobId = System.getenv("RAY_JOB_ID");
     String rayAddress = System.getProperty("ray.address");
     if (jobId != null && rayAddress != null) {
-      try (FileWriter writer = new FileWriter(logDir + "/java-worker-" + jobId + "-" + pid + ".log")) {
-        writer.write(":job_id:" + jobId + "\n");
+      String prefix = System.getProperty("ray.logging.file-prefix", "java-worker");
+      if ("java-worker".equals(prefix)) {
+        try (FileWriter writer = new FileWriter(logDir + "/" + prefix + "-" +
+                jobId + "-" + pid + ".log")) {
+          writer.write(":job_id:" + jobId + "\n");
```
carsonwang, line 79: Can you please comment why we write the jobId here?

carsonwang (commented):
Thanks @zjf2012 . can you please add some comments in the code before we merging this?

core/raydp-main/src/main/java/org/apache/spark/raydp/SparkOnRayConfigs.java
zjf2012: done

core/agent/src/main/java/org/apache/spark/raydp/Agent.java
zjf2012: done

core/raydp-main/src/main/java/org/apache/spark/raydp/SparkOnRayConfigs.java
carsonwang: can you fix the style to pass the CI？

zjf2012: just fixed

PR #329, opened 2023-04-04
Fix document since PR#313

pang-wu:
@carsonwang @kira-lin I forgot to update the doc since the merge of this PR: https://github.com/oap-project/raydp/pull/313


kira-lin:
Thanks

PR #333, opened 2023-04-18
[raydp-332] Fix security issue of protobuf < 3.19.5

zjf2012:
There is a security issue report, https://github.com/oap-project/raydp/security/dependabot/6.

```
Package protobuf

Affected versions >= 3.19.0, < 3.19.5

Patched version 3.19.5

 protobuf-cpp and protobuf-python have potential Denial of Service issue
```

To fix it, we limit protobuf > 3.19.5 and <= 3.20.3

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -102,7 +102,7 @@ try:
         "ray >= 2.1.0",
         "pyspark >= 3.1.1, <= 3.3.2",
         "netifaces",
-        "protobuf >= 3.15.3, <= 3.20.3, != 3.19.5"
+        "protobuf > 3.19.5, <= 3.20.3"
     ]
 
     _packages = find_packages()


zjf2012:
@carsonwang as tested with ray 2.1 - 2.3, it worked though there is a warning like below,

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
grpcio-tools 1.51.3 requires protobuf<5.0dev,>=4.21.6, but you have protobuf 3.20.3 which is incompatible.
```

kira-lin:
Is grpcio-tools used in our MPI test? Maybe we should downgrade its version as well?

zjf2012:
> Is grpcio-tools used in our MPI test? Maybe we should downgrade its version as well?
It's dep of ray
`
Requirement already satisfied: grpcio>=1.32.0 in /home/jiafu/anaconda3/envs/ray2.1/lib/python3.9/site-packages (from ray>=2.1.0->raydp==1.6.0.dev0) (1.51.3)`

zjf2012:
@kira-lin  @carsonwang do you have more comments?

kira-lin:
LGTM

kira-lin:
Ray's requirement for protobuf is "protobuf >= 3.15.3, != 3.19.5", why are we limiting it <=3.20.3?

zjf2012:
> grpcio-tools 1.51.3 requires protobuf<5.0dev,>=4.21.6,

It's due to grpcio-tools install protobuf-4.22.* which doesn't work well with raydp.

kira-lin:
Thanks, merged

PR #335, opened 2023-04-18
[raydp-334] fix the workaround of working with ray 2.3.[0-1]

zjf2012:
we need to prepend jobid to hardcoded prefix log file, java-worker-*.log instead of getting prefix from the "ray.logging.file-prefix" system property since ray hasn't applied the fix (https://github.com/ray-project/ray/pull/33665) yet.


zjf2012:
@kira-lin @carsonwang please help review.

kira-lin:
Why revert the change? It works both now and after that PR merged, right?

zjf2012:
> Why revert the change? It works both now and after that PR merged, right?

It doesn't work as expected with ray 2.3.1 for now since ray still uses 'java-worker' as prefix even though we've set and get system property like below,

`String prefix = System.getProperty("ray.logging.file-prefix", "java-worker");`

Thus, we need to prepend "jobid" to java-worker*.log directly until the ray PR merged.

PR #354, opened 2023-06-21
Update RayDP to be compatible with Ray 2.5.0

kira-lin:


kira-lin:
We cannot support Ray 2.5.0 as Ray Data batcher does not consider block type `bytes`. Even if we push a fix to Ray now, Ray 2.5.0 won't be updated. After our arrow serialization PR is merged, this problem will be fixed.

PR #362, opened 2023-07-07
[Follow-up] Fix log4j and some dependencies for `shims-common`

Deegue:
1. Exclude log4j:log4j in Spark 3.2.2
2. Fix missed in `shims-common`
3. Remove redirectLog() which is outdated

### DIFF
--- a/core/raydp-main/pom.xml
+++ b/core/raydp-main/pom.xml
@@ -61,6 +61,10 @@
           <groupId>org.apache.ivy</groupId>
           <artifactId>ivy</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>log4j</groupId>
+          <artifactId>log4j</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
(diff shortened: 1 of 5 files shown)


core/raydp-main/pom.xml
```diff
@@ -61,6 +61,10 @@
           <groupId>org.apache.ivy</groupId>
           <artifactId>ivy</artifactId>
         </exclusion>
+        <exclusion>
```
pang-wu, line 64: Spark is a provided dependency here, we shouldn't need to exclude as the class file won't be packaged

Deegue: Thanks for your review! Adding the exclusion here is to resolve security issues scanned by Synk before 1.6 release.

Just as you said, it won't affect users no matter which Spark version was used.

pang-wu: ah gotcha!

kira-lin:
I think core/pom.xml also needs an update, right? All poms using spark version 3.2.2 needs the update @Deegue

Deegue:
> I think core/pom.xml also needs an update, right? All poms using spark version 3.2.2 needs the update @Deegue

Good catch! Added.

Issue #364, opened 2023-07-10
Executor lost triggered by JVM OOM can't recover and cause subsequent steps stuck


pang-wu:
We found that if during execution the task get `OutOfMemoryError` and executor was lost, the Spark driver won't request new executors to backfill the ones killed by OOM. The following step could reproduce this error:

1. Create a dataset (could be a table, or parquet file) with large column which can't fit into your spark job's memory.
2. Start a small cluster, say with 3 executors with small memory that you know will OOM when persist the rdd. Load the data and persist:
```python
from pyspark import StorageLevel

df = spark.sql("select * from some_big_table")
rdd = df.rdd
rdd.persist(StorageLevel.MEMORY_AND_DISK)
rdd.count()
```

You should see errors similar to following:
```
23/07/10 04:54:29 WARN TaskSetManager: Lost task 1.0 in stage 8.0 (TID 105) (10.191.8.91 executor 6): java.lang.OutOfMemoryError: Java heap space
	at org.apache.parquet.bytes.BytesInput$StreamBytesInput.toByteArray(BytesInput.java:285)
	at org.apache.parquet.bytes.BytesInput.toByteBuffer(BytesInput.java:237)
...

23/07/10 04:54:29 ERROR TaskSchedulerImpl: Lost executor 6 on 10.191.8.91: Remote RPC client disassociated. Likely due to containers exceeding thresholds, or network issues. Check driver logs for WARN messages.

....
23/07/10 04:54:29 WARN BlockManagerMasterEndpoint: No more replicas available for rdd_13_6 !
23/07/10 04:54:29 WARN BlockManagerMasterEndpoint: No more replicas available for rdd_13_7 !
23/07/10 04:54:29 WARN BlockManagerMasterEndpoint: No more replicas available for rdd_13_8 !
```

Cluster will stuck in a bad state where it shows there are active tasks but all executors dead (if OOM happens on all executor, otherwise there would be executor without task scheduled alive):

<img width="1708" alt="Screenshot 2023-07-09 at 10 05 52 PM" src="https://github.com/oap-project/raydp/assets/104795337/6131c59f-5444-46e4-9a74-15051d115142">

This problem will impact some of the fault tolerant feature like `raydp.spark.from_spark_recoverable(df)` as OOM could happen inside the call -- this is how we found a way to reproduce this issue.

We can reproduce this using the following setup:
RayDP build from master branch 
Spark 3.3.2
Ray 2.4.0/2.3.1

@carsonwang

[...]

PR #389, opened 2023-11-08
Recovery for lost executors 364

KiranP-d11:
Pull request for the bug described in the issue https://github.com/oap-project/raydp/issues/364.

### DIFF
--- a/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
+++ b/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
@@ -131,6 +131,7 @@ class RayAppMaster(host: String,
 
     override def receive: PartialFunction[Any, Unit] = {
       case RegisterApplication(appDescription: ApplicationDescription, driver: RpcEndpointRef) =>
+
         logInfo("Registering app " + appDescription.name)
         val javaOpts = appendActorClasspath(appDescription.command.javaOpts)
         val newCommand = appDescription.command.withNewJavaOpts(javaOpts)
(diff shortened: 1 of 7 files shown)


core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
```diff
@@ -142,6 +142,37 @@ class RayAppMaster(host: String,
...
+      /**
+       * On receiving the event, we check if any executors are lost and mark them in the appInfo.
+       */
+      case MarkLostExecutors(appDescription: ApplicationDescription) =>
+        logDebug("Received an event for MarkLostExecutors")
+        val allExecutors = appInfo.getAllExecutorIds()
+        logDebug(s"Checking if any of the executor handlers are unreachable for ${allExecutors.mkString(", ")}")
+        allExecutors.foreach { executorId =>
+          val handlerOpt = appInfo.getExecutorHandler(executorId)
+          if (handlerOpt.isEmpty) {
```
kira-lin, line 154: This is not likely to happen, right?

kira-lin (commented):
Thanks for your PR!
I have a concern, though, which is how will this interact with fault tolerance mode? In fault tolerance mode, if an executor(actor) failed, it will get restarted and re-register. I think this pr does pretty much the same thing, but in Spark. If users turn on fault tolerant mode, will executors be requested multiple times?

Maybe I forgot the context, but why does fault tolerant mode failed to meet your need?

kira-lin:
Please also fix the style lint errors, thanks

KiranP-d11:
Correct me if I am wrong, but the fault tolerance mode is only for converting a Spark DataFrame to a Ray dataset. If the use case is to just run a Spark job and write the output to a destination (like S3), then enabling fault tolerance mode has no effect.

That being said, I understand that when the executor actors die, either because of OOM issues or node failures, Ray would restart the actors, and the restarted executors should be added back to Spark. However, currently, this is not happening. When an executor dies, it gets restarted, but it immediately dies before adding itself to the Spark application. This results in the Spark application getting stuck, where it indefinitely waits for the executors to come up without doing anything.

I will try to debug why this issue is happening and will attempt to raise a PR for it (in which case this current PR with changes in spark is not needed.)

kira-lin:
> However, currently, this is not happening. When an executor dies, it gets restarted, but it immediately dies before adding itself to the Spark application. This results in the Spark application getting stuck, where it indefinitely waits for the executors to come up without doing anything.

Do you mean even when fault tolerant mode is on, executors are not adding back as expected? I see, this might be some bug. 

We add fault tolerance mode to recover dataframes, indeed, and it introduces some behavior which might not be wanted. We are happy to have this feature, but we should make sure this can be turned off when fault tolerance mode is on.

KiranP-d11:
> Do you mean even when fault tolerant mode is on, executors are not adding back as expected? I see, this might be some bug.

Yes, even when the fault tolerance mode is on, executors are not adding back.

kira-lin:
I see.
> We are happy to have this feature, but we should make sure this can be turned off when fault tolerance mode is on.

Is it possible to check if fault tolerance mode is on, and if so disable the periodic sent RPC?

@pang-wu , does this solve your issue？

KiranP-d11:
>Is it possible to check if fault tolerance mode is on, and if so disable the periodic sent RPC?

I will check and get back on this.

pang-wu:
@kira-lin I need to check, but I am very exited about this feature -- it solve a big problem in our production. However I have a some concern on turn this feature off automatically, if possible, I would rather have an option for user to turn it off. The reasons are below:

> Is it possible to check if fault tolerance mode is on, and if so disable the periodic sent RPC?

Let's say if fault tolerance mode is on, but I am not converting Spark dataframe to Ray dataset. Will RayDP still recovers failed executors?
Also, we have other usecases where Spark context is not initialized using Python code, but using Spark submit/raydp submimt. In this case, we don't have a way to turn fault tolerance on event it can solve the problem..?

kira-lin:
> Will RayDP still recovers failed executors?

With fault tolerance mode on, we basically turn on Ray's failed actor restart, so yes, it should recover the executors by design.

> Also, we have other usecases where Spark context is not initialized using Python code, but using Spark submit/raydp submimt. In this case, we don't have a way to turn fault tolerance on event it can solve the problem..?

That's right.

KiranP-d11:
@kira-lin I was able to debug and fix the issue with ray's fault tolerance.
I have raised another [PR](https://github.com/oap-project/raydp/pull/391) for this.

Closing this PR as its is not required.

PR #391, opened 2023-11-29
Changes for executor actor recovery using ray's fault tolerance.

KiranP-d11:
Pull request for the bug described in the issue https://github.com/oap-project/raydp/issues/364.

This is the fix for ray executors not recovering from OOM and other failures.
The issue is because of the race condition:

```
 - Executor E1 dies lets say because of OOM
 - We try to kill it by firing stop call on E1 actor
 - Since the actor is not available, the stop task fails for E1
 - In the mean while, ray brings up the lost executor E1
 - The failed task (stop task) gets retried as there are task retries configured.
 - The stop task gets fired on the new executor which got recovered
 - The Recovered executor exits with status as user intended exit.
```


KiranP-d11:
@kira-lin Merged the master branch and CI checks are passing.

PR #392, opened 2023-11-30
Update raydp-submit

gptbert:
Traceback (most recent call last):
  File "<string>", line 54, in <module>
NameError: name 'os' is not defined


kira-lin:
thanks

PR #396, opened 2023-12-15
dynamic allocation executor pending for addition flood fix

raviranak:
## Overview 
Issue : 
Currently with dynamic allocation for executors there are too many executors pending to be added while the actual number of max executor is far less 

Solution:
This will check with dynamic auto scale no additional pending executor actor added more than max executors count
as this result in executor even running after job completion


raviranak:
@kira-lin Can you please take a look at this

kira-lin:
Sorry for the late reply. I have a question: do you mean that the executors killed by Spark due to dynamic resource allocation will try to restart?

raviranak:
> Sorry for the late reply. I have a question: do you mean that the executors killed by Spark due to dynamic resource allocation will try to restart?

With dynamic resource allocation the amount of pending executor creation is way greater than actual number of executor , so even after completion of spark job the executor is alive for much longer

rishabh-dream11:
@kira-lin Can you help with the review and merge?

kira-lin (approved):

PR #397, opened 2023-12-21
Support Spark 3.4.2, lift pyspark dependency range

pang-wu:


pang-wu:
@kira-lin @carsonwang Do you mind take a look?

kira-lin:
Thanks

Issue #380, opened 2023-09-21
Runtime env_vars cannot be passed down to executors


Deegue:
When we use `ray.init(runtime_env={"env_vars":"xxx"})`, executors cannot get "env_vars" while it is critical for executors in many scenarios.

It is because `RayAppMaster` will raise a new job which is separated from the worker pod of `RayDPSparkMaster`.

g-clef:
I'm running in to this also. Would like to be able to run experiments with different dependent libraries (and send them to the workers with an env_var) without having to change the base image of my ray cluster.

ujjawal-khare-27:
If you are having ray clusters on k8s, you can introduce these variables in pods env.

PR #420, opened 2025-01-24
modify ray_cluster master settings to accept and pass on runtime_env #380

g-clef:
Looking at issue #380 , it looks like the solution is fairly straightforward - since `_set_up_master` in ray_cluster.py already accepted a `kwargs` parameter, but didn't use it, I renamed that to `runtime_env`, and passed it on to the RayDPSparkMaster options call.


carsonwang:
Thanks for the PR. Can we add a test for this?

pang-wu (commented):
Can you elaborate a bit more on what problem you want to solve here?
If we want to bring runtime env to the ray actor, we probably also want to do that for the executors as well?

PR #438, opened 2025-10-12
CI switch to jdk17, pin click, support Spark 3.5.5-7

pang-wu:


carsonwang:
@pang-wu thanks for the work! It seems not all examples were updated with the JDK17 options, can you please update all of them?

pang-wu:
> @pang-wu thanks for the work! It seems not all examples were updated with the JDK17 options, can you please update all of them?

@carsonwang Done

PR #447, opened 2025-11-21
tf/estimator.py: only write checkpoint in rank0

pang-wu:


python/raydp/tf/estimator.py
```diff
@@ -188,9 +188,14 @@ def train_func(config):
...
                 results.append(test_history)
+
+        # Only save checkpoint from the chief worker to avoid race conditions.
+        # However, we need to call save on all workers to avoid deadlock.
         with tempfile.TemporaryDirectory() as temp_checkpoint_dir:
             multi_worker_model.save(temp_checkpoint_dir, save_format="tf")
-            checkpoint = Checkpoint.from_directory(temp_checkpoint_dir)
+            checkpoint = None
+            if session.get_world_rank() == 0:
+                checkpoint = Checkpoint.from_directory(temp_checkpoint_dir)
```
carsonwang: In the torch estimator, we only call save in rank 0. But why will it cause deaklock here and need to call on all workers?

carsonwang: If we follow the document you mentioned, saving checkpoint should also be only  called in rank0, right? But the currently code saves checkpoint for every worker, it just loads and reports checkpoint in rank0.

pang-wu: You probably want to refer to this (and yes I think even the checkpoint doc in Ray doesn't cover the edge case from Keras): 
https://www.tensorflow.org/tutorials/distribute/multi_worker_with_keras#model_saving_and_loading
https://www.tensorflow.org/tutorials/distribute/multi_worker_with_keras#checkpoint_saving_and_restoring

In keras, `Model.save` involve collective operation, which it will wait all workers call the same `save` method otherwise the chief worker won't return. 

TF’s recommended pattern in multi-worker is: all workers execute the saving code path, but only the chief actually writes the final artifact (via options or role-checks). In ray, the `report` method is what triggers the final checkpoint write.

carsonwang (approved):

PR #448, opened 2025-11-22
Allow assign memory to RayDPSparkMaster actor

pang-wu:


pang-wu:
> Looks good. Do you also want to update the document?

Done -- I left the docker image repo stays as-is.

PR #450, opened 2025-12-08
Support Spark 4.x

pang-wu:
This PR adapt raydp with Spark 4.x but leave the following work for future improvement:
1. Support tensorflow 2.16+ (see https://keras.io/getting_started/#tensorflow--keras-2-backwards-compatibility) and numpy 2.x
2. Support python 3.12 - we deprecated Python 3.9 because it is no longer supported by Spark. Need to modernize python build system.
3. Deprecate Ray AIR.

To make the tests pass, the PR is based on https://github.com/ray-project/raydp/pull/458. Once PR#458 is merged this PR should rebase again.


rexminnis (commented):
Thanks for putting this together — the `CommandLineUtilsBridge` pattern and the `SparkSubmit` rework are clean solutions to the cross-version API drift. A few things I noticed:

1. **Bug:** `spark340/SparkSqlUtils.toArrowRDD` has infinite recursion (see inline comment)
2. **Java target:** `maven.compiler.source` is still `1.8` — worth bumping to 17?
3. **Spark version:** `spark410.version` targets 4.1.0 — consider 4.1.1 (current release)

Happy to help with testing or any of the shim work. I have a working Spark 4.1.1 setup locally and have been validating the Arrow conversion paths end-to-end.

carsonwang:
@pang-wu Great work! The changes to support spark 4.x look good. Please just do a cleanup (remove some outdated code from PR458) and then I think it is ready to merge.

pang-wu:
Thanks @carsonwang! Cleanup is in 539bf7d — removed the legacy non-recoverable path from `ObjectStoreWriter.scala` (the instance `ObjectStoreWriter` class with `dfToId`/`getRandomRef`/`clean`, the deprecated `fromSparkRDD` method, the `ObjectRefHolder` object, and the now-unused imports). PTAL when you get a chance.

python/setup.py
```diff
@@ -97,11 +97,11 @@ def run(self):
     install_requires = [
         "numpy",
         "packaging",
-        "pandas >= 1.1.4",
+        "pandas >= 2.2.0, < 3.0.0",
         "psutil",
         "pyarrow >= 4.0.1",
         "ray >= 2.37.0",
-        "pyspark >= 3.1.1, <=3.5.7",
+        "pyspark >= 4.0.0",
```
carsonwang: Add upper bound?

core/shims/spark410/src/main/scala/org/apache/spark/sql/SparkSqlUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+    val schema = DataType.fromJson(schemaString).asInstanceOf[StructType]
+    val timeZoneId = session.sessionState.conf.sessionLocalTimeZone
+    val internalRowRdd = arrowBatchRDD.rdd.mapPartitions { iter =>
+      val context = TaskContext.get()
+      ArrowConverters.fromBatchIterator(
+        arrowBatchIter = iter,
+        schema = schema,
+        timeZoneId = timeZoneId,
+        errorOnDuplicatedFieldNames = false,
+        largeVarTypes = false,
```
carsonwang: This should honor the session config sparkSession.sessionState.conf.arrowUseLargeVarTypes too.

core/shims/spark400/src/main/scala/org/apache/spark/sql/SparkSqlUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+    val schema = DataType.fromJson(schemaString).asInstanceOf[StructType]
+    val timeZoneId = session.sessionState.conf.sessionLocalTimeZone
+    val internalRowRdd = arrowBatchRDD.rdd.mapPartitions { iter =>
+      val context = TaskContext.get()
+      ArrowConverters.fromBatchIterator(
+        arrowBatchIter = iter,
+        schema = schema,
+        timeZoneId = timeZoneId,
+        errorOnDuplicatedFieldNames = false,
+        largeVarTypes = false,
```
carsonwang: This should honor the session config sparkSession.sessionState.conf.arrowUseLargeVarTypes too.

python/setup.py
pang-wu: Done in 6bb6ff2 — capped at `< 5.0.0` (next major; Spark generally keeps API stable across minor versions, but a major bump may be breaking).

core/shims/spark400/src/main/scala/org/apache/spark/sql/SparkSqlUtils.scala
pang-wu: Done in 6bb6ff2 — `toDataFrame` now reads `sparkSession.sessionState.conf.arrowUseLargeVarTypes` (captured outside the closure like `timeZoneId`).

core/shims/spark410/src/main/scala/org/apache/spark/sql/SparkSqlUtils.scala
pang-wu: Done in 6bb6ff2 — `toDataFrame` now reads `sparkSession.sessionState.conf.arrowUseLargeVarTypes` (captured outside the closure like `timeZoneId`).

carsonwang (approved):
Thanks for the all the efforts!

PR #458, opened 2026-02-01
Handle gc prematurely evict cached RDD

pang-wu:
There is a race condition in current recoverable fetch:

## Problem

prepareRecoverableRDD in [ObjectStoreWriter.scala](https://github.com/ray-project/raydp/pull/core/raydp-main/src/main/scala/org/apache/spark/sql/raydp/ObjectStoreWriter.scala) creates and caches an Arrow RDD, but only returns its integer ID. After the method returns, the RDD has no strong JVM reference. Spark's ContextCleaner can GC the RDD and remove all cached blocks before Ray tasks fetch them.

The recache fallback (in [RayDPDriverAgent.scala](https://github.com/ray-project/raydp/pull/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayDPDriverAgent.scala), line 63) looks up the RDD via spark.getPersistentRDDs -- but once the RDD is cleaned, it's no longer in that map, so recache is silently a no-op.

Screen Shot 2026-02-16 at 11 38 19 PM

### 1. Pin recoverable RDDs in `ObjectStoreWriter` (Scala)

**File:** `core/raydp-main/src/main/scala/org/apache/spark/sql/raydp/ObjectStoreWriter.scala`

- Add a `ConcurrentHashMap[Int, RDD[_]]` to `object ObjectStoreWriter` to hold strong references keyed by `rdd.id`
- In `prepareRecoverableRDD`, after `rdd.count()`, store: `recoverableRDDs.put(rdd.id, rdd)`
- Add a `releaseRecoverableRDD(rddId: Int)` method that removes the entry and calls `rdd.unpersist()`

This mirrors how `dfToId` already keeps DataFrame references alive for the non-recoverable path (line 61).

### 2. Expose cleanup from Python (Python)

**File:** `python/raydp/spark/dataset.py`

- No auto-release in `from_spark_recoverable` (the recoverable pipeline needs the cache alive for Ray task retries)
- The pinned RDD will be released automatically when SparkContext stops
- Optionally add a public helper so users can explicitly release when done:

```python
def release_recoverable_rdd(spark_context, rdd_id):
    sc = spark_context
    sc._jvm.org.apache.spark.sql.raydp.ObjectStoreWriter.releaseRecoverableRDD(rdd_id)
```

### DIFF
--- a/python/raydp/spark/dataset.py
+++ b/python/raydp/spark/dataset.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 import logging
 import uuid
-from typing import Callable, List, Optional, Union
+from typing import Callable, Dict, List, Optional, Union
 from dataclasses import dataclass
 
 from packaging import version
@@ -43,6 +43,8 @@ from raydp.utils import parse_memory_size
 
 
 logger = logging.getLogger(__name__)
+
+_recoverable_rdd_ids: Dict[sql.DataFrame, int] = {}
 
 def _enable_load_code_from_local() -> None:
     """Enable Ray cross-language support via internal API (driver only)."""
@@ -171,6 +173,7 @@ def from_spark_recoverable(df: sql.DataFrame,
                            storage_level: StorageLevel = StorageLevel.MEMORY_AND_DISK,
                            parallelism: Optional[int] = None):
     """Recoverable Spark->Ray conversion that survives executor loss."""
+    original_df = df
     num_part = df.rdd.getNumPartitions()
     if parallelism is not None:
         if parallelism != num_part:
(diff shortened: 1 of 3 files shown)


pang-wu:
> Thanks for the fix! In Spark, blockManager.get(blockId) already tries remote fetch if the block isn't local. If that worked in RayDP, the second blockManager.get() after recache would transparently succeed regardless of which executor holds the block, so no forwarding is needed. Could you please confirm this works or not? In the test, is it possible to test the scenario with recache?

Thanks for the review @carsonwang. I removed the forwarding -- this is just trying to be defensive and I actually can't reproduce that recover path after fixing the GC issue.

python/raydp/spark/dataset.py
```diff
@@ -212,6 +217,20 @@ def from_spark_recoverable(df: sql.DataFrame,
 
     return from_arrow_refs(refs)
 
+def release_recoverable_rdd(spark: sql.SparkSession, rdd_id: int) -> None:
```
carsonwang: This is a public API. How can user get the rdd_id to release? we might need update from_spark_recoverable to return the id?

carsonwang (approved):
LGTM

PR #463, opened 2026-02-17
Fix TaskContext leak and resource cleanup in                
   getRDDPartition

rexminnis:
## Summary                                                                                                                                            
   -  sets a dummy  but never unsets it. Since executor actors reuse threads (), a stale  leaks into subsequent      
   operations on the same thread, which can cause subtle issues with metrics, accumulators, and task-local state.
   - If an exception occurs mid-method, neither the / nor the  read locks are cleaned up.
   - Wraps the method body in  to guarantee:
     -  read locks are released via 
     -  is unset via 
     -  and  are closed

   ## Test plan
   - [x] Verify core compiles cleanly against all shim targets (3.2.2, 3.3.0, 3.4.0, 3.5.0)
   - [ ] Run  round-trip conversion
   - [ ] Confirm no resource leak warnings in executor logs after repeated conversions.


pang-wu:
Thanks for the PR -- I will address this in https://github.com/ray-project/raydp/pull/450/changes/1fa38467600b9465c1e3257f95141ff756264310

PR #467, opened 2026-03-10
Fix RayDPDriverAgent Rpc bind

pang-wu:
When `fault_tolerant_mode=True` (the default), `connect_spark_driver_to_ray()` calls `RayAppMaster.setProperties(jvm_properties)` which loads the **master actor's** Ray config into the driver JVM. This sets `ray.node-ip` to the master's IP. Then `ObjectStoreWriter.connectToRay()` creates a `RayDPDriverAgent`, which calls `RayConfig.create().nodeIp` to get the bind address -- but this now returns the **master's** IP, not the driver's local IP.

The call chain:

```
init_spark(fault_tolerant_mode=True)
  -> connect_spark_driver_to_ray()
    -> RayAppMaster.setProperties(master_ray_configs)    // sets ray.node-ip = master's IP
    -> ObjectStoreWriter.connectToRay()
      -> new RayDPDriverAgent()
        -> RayConfig.create().nodeIp                      // returns master's IP (WRONG)
        -> RpcEnv.create(..., host=masterIp, ...)         // BindException!
```

### DIFF
--- a/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayDPDriverAgent.scala
+++ b/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayDPDriverAgent.scala
@@ -17,8 +17,6 @@
 
 package org.apache.spark.deploy.raydp
 
-import io.ray.runtime.config.RayConfig
-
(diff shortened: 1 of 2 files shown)


carsonwang (approved):
Thanks for the fix!

PR #475, opened 2026-05-17
fix(executor): switch process working dir via native chdir

my-vegetable-has-exploded:
## Motivation
The RayDP executor process CWD remains the container default directory (e.g., / or /opt) rather than the executor's workingDir. This causes Spark distributed files (--files) and archives (--archives) to be extracted to the wrong location, making it impossible for executor code to find them at the expected paths.
## Approach
- JNA native `chdir` call: Add JNA dependency to `pom.xml`, define a `LibC` interface to load the `chdir()` syscall from libc. After `setUserDir()` (which only changes the user.dir system property), add `switchProcessWorkingDirBestEffort()` to actually switch the process-level CWD
- Align `SparkEnv.driverTmpDir`: Change `driverTmpDir` from `workingDir/_tmp` to `workingDir` itself, so the root directory for Spark distributed files matches the executor's `workingDir`, and files/archives are extracted to the correct path
- Fault tolerance: `chdir` failure only logs a warning and does not interrupt executor startup; reads `cwd` before and after the switch for diagnostic logging

### DIFF
--- a/core/raydp-main/src/main/scala/org/apache/spark/executor/RayDPExecutor.scala
+++ b/core/raydp-main/src/main/scala/org/apache/spark/executor/RayDPExecutor.scala
@@ -25,8 +25,8 @@ import java.util.concurrent.atomic.AtomicBoolean
 
 import scala.reflect.classTag
 
-import com.sun.jna.{Library, Native}
 import com.intel.raydp.shims.SparkShimLoader
+import com.sun.jna.{Library, Native}
 import io.ray.api.Ray
 import io.ray.runtime.config.RayConfig
 import org.apache.arrow.vector.ipc.{ArrowStreamWriter, WriteChannel}
--- a/core/raydp-main/src/main/scala/org/apache/spark/executor/RayDPExecutor.scala
+++ b/core/raydp-main/src/main/scala/org/apache/spark/executor/RayDPExecutor.scala
@@ -19,11 +19,13 @@ package org.apache.spark.executor
 
 import java.io.{ByteArrayOutputStream, File}
 import java.nio.channels.Channels
+import java.nio.file.Files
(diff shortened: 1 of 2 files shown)


core/raydp-main/src/main/scala/org/apache/spark/executor/RayDPExecutor.scala
```diff
@@ -238,7 +283,10 @@ class RayDPExecutor(
       val workerTmpDir = new File(workingDir, "_tmp")
       workerTmpDir.mkdir()
       assert(workerTmpDir.exists() && workerTmpDir.isDirectory)
-      SparkEnv.get.driverTmpDir = Some(workerTmpDir.getAbsolutePath)
+      // Keep Spark's distributed file/archive root aligned with executor workingDir.
+      SparkEnv.get.driverTmpDir = Some(workingDir.getAbsolutePath)
```
pang-wu, line 287: will only do this solve the problem?

core/raydp-main/src/main/scala/org/apache/spark/executor/RayDPExecutor.scala
```diff
@@ -188,6 +191,48 @@ class RayDPExecutor(
...
+        s"target is ${targetDir}")
+    } catch {
+      case e: Throwable =>
+        logWarning(s"Failed to switch executor process cwd from ${beforeCwd} to ${targetDir}", e)
+    }
+  }
+
+  private def getProcessWorkingDir: String = {
+    try {
+      val procCwd = Paths.get("/proc/self/cwd")
```
pang-wu, line 217: This is linux specific.

pang-wu:
Can we add some tests to verify the fix work?

PR #481, opened 2026-05-18
fix(driver): force JVM exit to prevent hanging RayJobs

my-vegetable-has-exploded:
## Motivation

In Spark on Ray scenarios, the driver logic may finish but the JVM cannot exit naturally because non-daemon threads (e.g. Hudi Embedded Timeline Server, connection pools, etc.) keep it alive. KubeRay RayJob relies on the entrypoint process exiting to determine job completion, so this causes RayJobs to stay in `RUNNING` state indefinitely.

## Approach

1. **`JvmExitGuard`** — a singleton that, when armed, starts a daemon countdown thread. If the JVM hasn't exited within the configured timeout, it forcibly calls `System.exit()`. If `System.exit()` throws `IllegalStateException` (JVM already in its shutdown sequence), it falls back to `Runtime.getRuntime().halt()` which bypasses all shutdown hooks and terminates immediately.

2. **Unified exit code constants** — `EXIT_SUCCESS=0`, `EXIT_APP_FAILED=1`, `EXIT_KILLED=143`. Replace all hardcoded exit codes in `DriverExitState` and `ApplicationInfo` with these constants.

3. **Arm points**:
   - `finalizeDriverTermination()` — called on normal exit, failure, and exit function redirect
   - `stop(KILLED)` path in `RayCoarseGrainedSchedulerBackend`

4. **`raydp-submit` integration** — forward `RAYDP_JVM_EXIT_TIMEOUT` environment variable as `-Draydp.jvm.exit.timeout` JVM property.

### DIFF
--- a/bin/raydp-submit
+++ b/bin/raydp-submit
@@ -143,6 +143,12 @@ added_args+=("spark.driver.host=$RAY_DRIVER_NODE_IP")
 added_args+=("--conf")
 added_args+=("spark.driver.bindAddress=$RAY_DRIVER_NODE_IP")
 
+# JvmExitGuard: force JVM exit after Spark app ends if non-daemon threads keep it alive.
+# Set RAYDP_JVM_EXIT_TIMEOUT env var to override the default timeout (300s) or disable (0).
+if [ -n "$RAYDP_JVM_EXIT_TIMEOUT" ]; then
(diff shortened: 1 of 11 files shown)


my-vegetable-has-exploded:
wait for #480

PR #487, opened 2026-07-23
fix(RayAppMaster): refactor executor state in RayAppMaster.

my-vegetable-has-exploded:
## What does this PR do?

The old code counted Spark executor generations instead of logical Ray actors. A restarted executor was recorded in both `executors` and `restartedExecutors`, while stale restart mappings survived executor removal. This allowed `registeredExecutors` to report a deficit while the restart-based limit check simultaneously reported that the application was full. `RequestExecutors` would then acknowledge the target without creating any actor. Tracking slots by the stable Ray `actorId` removes this contradiction: a restart changes the Spark executor generation, not the number of logical actors.

Two lifecycle details are essential to make that model correct. First, the actor-slot registry must retain an `ActorHandle` across executor disconnects; otherwise Ray's reference can be released and the non-detached actor terminated while RayDP still counts its slot. Second, RayAppMaster must reconcile after `KillExecutors`: Spark treats the requested total as persistent and expects the cluster manager, as YARN does, to replace missing executors without another identical request from the driver.

### Problem

RayDP previously conflated two different identities:

- A Ray actor has a stable identity and keeps the same named `actorId` across restarts.
- A Spark executor gets a new `executorId` for every restarted generation.

The previous implementation used several generation-based states to approximate the number of executor actors:

- `registeredExecutors` to calculate how many executors were missing.
- `executors.size + restartedExecutors.size` to enforce the executor limit.
- `restartedExecutors` to retain the mapping from a new executor ID to its original actor ID.

These states have different lifecycles and can disagree. In particular, a restarted executor is present in both `executors` and `restartedExecutors`, so the same Ray actor is counted twice. Moreover, entries in `restartedExecutors` are not removed when the corresponding executor generation disconnects, allowing historical restarts to remain in executor accounting.

### Failure Scenario

The incorrect count follows directly from three asymmetric code paths:

1. When a restarted actor requests a new Spark executor ID, RayAppMaster calls `addPendingRegisterExecutor(newExecutorId, ...)`, which inserts the new generation into `appInfo.executors`. It then also writes `restartedExecutors(newExecutorId) = originalActorId`.
2. Every `newExecutorId` is unique. A later restart therefore adds another map entry instead of replacing the entry for the previous generation, even though both values point to the same original Ray actor.
3. On RPC disconnect, `ApplicationInfo.kill()` removes the current generation from `appInfo.executors`. It cannot remove the matching `restartedExecutors` entry because that map is owned separately by `RayAppMaster`, and neither the disconnect path nor `KillExecutors` performs such cleanup.

The capacity check nevertheless uses:

```text
appInfo.executors.size + restartedExecutors.size
```

This is wrong in two different phases. While a restarted generation is active, its executor ID is present in both maps and is counted twice. After it disconnects, it is removed from `executors` but remains in `restartedExecutors`, so it becomes a permanently counted historical generation.

For one logical actor with original `actorId = 7`, the state evolves as follows:

| Event | `appInfo.executors` | `restartedExecutors` | Count used by limit check | Logical actor slots |
| --- | --- | --- | ---: | ---: |
| Initial generation | `{7}` | `{}` | 1 | 1 |
| Restart as executor 1001 | `{1001}` | `{1001 -> 7}` | 2 | 1 |
| Executor 1001 disconnects | `{}` | `{1001 -> 7}` | 1 | 1 |
| Restart as executor 1002 | `{1002}` | `{1001 -> 7, 1002 -> 7}` | 3 | 1 |

Thus, each restart adds another capacity unit to `restartedExecutors`, although the number of Ray actor slots never changes. Disconnecting the generation does not undo that increase.

The restart-registration path does not call `requestNewExecutor()`, so it bypasses the maximum check that later consumes this inflated count. Each disconnect also decrements `registeredExecutors`, making `remainingUnRegisteredExecutors > 0` true and admitting the next restart generation. The history can therefore grow to the configured maximum and beyond.

The key contradiction is:

> `registeredExecutors` says executors are missing, while `restartedExecutors` says the
> application is already at its limit.

Both cannot be valid measures of Ray executor capacity.

### Solution

Track logical Ray actor slots independently from Spark executor generations.

An actor slot is keyed by the original `actorId`:

- Creating a new Ray actor adds one actor slot.
- Restarting that actor creates a new Spark `executorId`, but does not add another slot.
- An executor disconnect removes only that Spark executor generation because Ray may restart the
  same actor.
- When Spark sends `KillExecutors`, RayDP removes the actor slot and requests actor shutdown.
- The Spark driver's requested total is stored as `desiredExecutors`.

These rules describe two stages that may occur in the same failure path. The AppMaster's initial RPC disconnect handling keeps the slot because Ray may recover the actor. If Spark subsequently decides to terminate that executor and sends `KillExecutors`, RayDP then removes the slot and stops the actor. Therefore slot removal is tied to the `KillExecutors` command, not only to an explicit scale-down reason.

Executor reconciliation then becomes:

```text
actorsToAdd = max(0, desiredExecutors - actorSlots)
```

This compares two values with matching semantics: the number of logical actors requested by Spark and the number of logical Ray actor slots currently owned by the application.

The implementation keeps an `actorId -> ActorHandle` registry as the actor-slot state and maps each Spark executor generation back to its stable actor ID. The restarted-executor mapping required by recoverable RDD lookup is derived from this state, but is no longer used for capacity accounting.

The actor-slot registry cannot continue to use the previous `executorIdToHandler` map. When an executor disconnects, `onDisconnected` calls `ApplicationInfo.kill(..., shutdownActor = false)` to remove that Spark executor generation while allowing Ray to restart its actor. However, `kill()` unconditionally executes `executorIdToHandler -= executorId`. Removing the executor ID also removes the map entry that retains the corresponding actor handle. If this is the last retained handle, the handle becomes unreachable and the actor can be terminated, which violates the expected restart semantics of `shutdownActor = false`.

The handle is not only a lookup convenience: it keeps Ray's reference to a non-detached actor alive. RayDP creates a named actor with `setName`, but does not give it `ActorLifetime.DETACHED`. In Ray 2.1, `NativeActorHandleReference.finalizeReferent()` calls `nativeRemoveActorHandleReference()` after a Java handle becomes unreachable. Losing the last retained handle can therefore reduce the distributed reference count to zero and terminate the actor. `setMaxRestarts(-1)` allows recovery from actor process failures; it does not make the actor independent of its owning reference.

The sharp failure sequence is:

```text
actorId = 23, current Spark executorId = 1001
executorIdToHandler(1001) = handle(23)

executor 1001 disconnects
  -> kill(1001, shutdownActor = false)
  -> actorSlots still contains 23, because Ray is expected to restart it
  -> executorIdToHandler removes 1001 unconditionally
  -> handle(23) becomes unreachable and is garbage-collected
  -> Ray removes the actor-handle reference and actor 23 can be terminated
```

RayDP is then left with a ghost slot: `actorSlots` says actor 23 exists, while the actor and its handle are gone. Because reconciliation sees the slot as present, it creates no replacement. A named actor is not sufficient protection here; naming makes lookup possible only while the actor is still alive.

For this reason, `f61f80e5e6e0daa125172573ecf4c8262aea2af1` replaces `executorIdToHandler` with two maps:

```text
executorId -> actorId
actorId    -> ActorHandle
```

The first map follows the lifetime of a Spark executor generation. The second is both the actor-slot registry and the strong-reference holder for the Ray actor. Removing a disconnected executor ID now deletes only the generation mapping. The stable actor handle remains until the actor slot itself is deliberately removed, so Ray can restart the actor and register a new Spark `executorId`.

### Reconcile Immediately after `KillExecutors`

Spark's total-executor request is a persistent desired-state contract, not a one-shot request to create a delta. `CoarseGrainedSchedulerBackend` explicitly assumes that the cluster manager will eventually fulfill an acknowledged total request when resources become available. Consequently, the Spark driver does not periodically resend the same target merely because an executor disappears; the resource-management side is responsible for retaining the target and reconciling its actual resources against it.

YARN demonstrates this contract directly. `YarnAllocator` stores the driver's total in `targetNumExecutorsPerResourceProfileId`. On every allocator heartbeat, it computes:

```text
missing = target - pending - running - starting
```

When a container exits, the running count decreases and the next heartbeat requests a replacement, without requiring the Spark driver to resend its unchanged target.

Before this commit, RayDP retained `desiredExecutors`, but reconciled only when it received `RequestExecutors`. Consider a target of 1000 and Spark calling `killAndReplaceExecutor(17)`. Spark sends `KillExecutors` with `adjustTargetNumExecutors = false`, so its desired target correctly stays at 1000. RayAppMaster removes actor slot 17 and acknowledges the kill, leaving 999 slots. No new `RequestExecutors(1000)` is required by Spark's contract, and RayAppMaster had neither a periodic allocator loop nor a reconciliation call on this state transition. The application could therefore remain at 999 indefinitely; repeated replacement kills could accumulate the deficit.

This commit invokes `reconcileExecutors()` immediately after processing `KillExecutors`:

- For a real scale-down, Spark first lowers the total request, so `desiredExecutors` already matches the reduced slot count and no replacement is created.
- For kill-and-replace or health-related removal, Spark keeps the total request unchanged, so the removed slot immediately appears as a deficit and RayDP creates a replacement.

`RequestExecutors` and `KillExecutors` are the two events that change RayAppMaster's target or actor slot count. Reconciling on both events gives RayDP the same desired-state behavior that Spark expects from resource managers such as YARN, without requiring a periodic target synchronization from the driver.

## Related Issue

Fixes #

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Test update
- [x] Refactoring
- [ ] Other

## How was this tested?

This fix has been tested in our production environment and works well with dynamic execution enabled and 1,000 executors. Before this fix, large-scale jobs would occasionally fail.

## Checklist

- [ ] I have added or updated tests if needed
- [ ] I have updated documentation if needed
- [ ] I have run relevant tests locally
- [ ] I have checked that this change is backward compatible

### DIFF
--- a/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
+++ b/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
@@ -58,8 +58,6 @@ private[spark] class ApplicationInfo(
   var coresGranted: Int = _
   var endTime: Long = _
   private var nextExecutorId: Int = _
-  // this only count those registered executors and minus removed executors
-  private var registeredExecutors: Int = 0
   // Desired executor target comes from the Spark driver. Actor handles track the Ray actor
   // slots AppMaster still owns, independent of transient Spark executor generations.
   private var desiredExecutors: Int = _
@@ -116,7 +114,6 @@ private[spark] class ApplicationInfo(
         false
       } else {
         executors(executorId).registered = true
-        registeredExecutors += 1
         true
       }
     } else {
@@ -140,9 +137,6 @@ private[spark] class ApplicationInfo(
   def kill(executorId: String, shutdownActor: Boolean): Boolean = {
     if (executors.contains(executorId)) {
       val exec = executors(executorId)
-      if (exec.registered) {
-        registeredExecutors -= 1
-      }
       removedExecutors += exec
       executors -= executorId
       executorIdToActorId -= executorId
@@ -171,14 +165,6 @@ private[spark] class ApplicationInfo(
   def getExecutorHandler(
       executorId: String): Option[ActorHandle[RayDPExecutor]] = {
     executorIdToActorId.get(executorId).flatMap(actorIdToHandle.get)
-  }
-
-  def remainingUnRegisteredExecutors(): Int = {
-    desc.numExecutors - registeredExecutors
-  }
-
-  def currentExecutors(): Int = {
-    registeredExecutors
   }
 
   def getNextExecutorId(): Int = {
--- a/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
+++ b/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
@@ -39,7 +39,13 @@ case class ExecutorDesc(
     memoryPerExecutorMB: Int,
     resources: Map[String, ResourceInformation]) {
   var registered: Boolean = false
+  var address: Option[RpcAddress] = None
 }
+
+private[spark] class ExecutorActorSlot(
+    var handle: ActorHandle[RayDPExecutor],
+    var currentExecutorId: Option[String],
+    var previousExecutorId: Option[String])
 
 private[spark] class ApplicationInfo(
     val startTime: Long,
@@ -52,8 +58,9 @@ private[spark] class ApplicationInfo(
   var state: ApplicationState.Value = _
   var executors: HashMap[String, ExecutorDesc] = _
   var addressToExecutorId: HashMap[RpcAddress, String] = _
+  // Resolves both current executor ids and the previous-generation tombstone to an actor slot.
   var executorIdToActorId: HashMap[String, String] = _
-  var actorIdToHandle: HashMap[String, ActorHandle[RayDPExecutor]] = _
+  var actorSlots: HashMap[String, ExecutorActorSlot] = _
   var removedExecutors: ArrayBuffer[ExecutorDesc] = _
   var coresGranted: Int = _
   var endTime: Long = _
@@ -69,7 +76,7 @@ private[spark] class ApplicationInfo(
     executors = new HashMap[String, ExecutorDesc]
     addressToExecutorId = new HashMap[RpcAddress, String]
     executorIdToActorId = new HashMap[String, String]
-    actorIdToHandle = new HashMap[String, ActorHandle[RayDPExecutor]]
+    actorSlots = new HashMap[String, ExecutorActorSlot]
     endTime = -1L
     nextExecutorId = 0
     desiredExecutors = desc.numExecutors
@@ -84,9 +91,13 @@ private[spark] class ApplicationInfo(
       memoryInMB: Int): Unit = {
     // Adding a pending executor also declares that its Ray actor slot is still active.
     // For restarted executors, actorId points back to the original named Ray actor.
-    actorIdToHandle(actorId) = handler
+    val slot = actorSlots.getOrElseUpdate(
+      actorId, new ExecutorActorSlot(handler, None, None))
+    slot.handle = handler
+    slot.currentExecutorId = Some(executorId)
     val desc = ExecutorDesc(executorId, actorId, cores, memoryInMB, null)
     executors(executorId) = desc
+    // Keep the previous generation mapping until the next disconnect or explicit actor shutdown.
     executorIdToActorId(executorId) = actorId
   }
 
@@ -95,7 +106,7 @@ private[spark] class ApplicationInfo(
   }
 
   def numActorsToAdd: Int = {
-    math.max(0, desiredExecutors - actorIdToHandle.size)
+    math.max(0, desiredExecutors - actorSlots.size)
   }
 
   // Compatibility view for ObjectStoreWriter: map restarted Spark executor ids back to
@@ -123,28 +134,33 @@ private[spark] class ApplicationInfo(
   }
 
   def markExecutorStarted(executorId: String, address: RpcAddress): Unit = {
-    addressToExecutorId(address) = executorId
-  }
-
-  def kill(address: RpcAddress, shutdownActor: Boolean): Boolean = {
-    if (addressToExecutorId.contains(address)) {
-      kill(addressToExecutorId(address), shutdownActor)
-    } else {
-      false
+    executors.get(executorId).foreach { exec =>
+      exec.address = Some(address)
+      addressToExecutorId(address) = executorId
     }
   }
 
+  def kill(address: RpcAddress, shutdownActor: Boolean): Boolean = {
+    addressToExecutorId.get(address).exists(kill(_, shutdownActor))
+  }
+
   def kill(executorId: String, shutdownActor: Boolean): Boolean = {
-    if (executors.contains(executorId)) {
-      val exec = executors(executorId)
-      removedExecutors += exec
-      executors -= executorId
-      executorIdToActorId -= executorId
-      coresGranted -= exec.cores
+    val actorIdOpt = executorIdToActorId.get(executorId)
+
+    actorIdOpt.foreach { actorId =>
       if (shutdownActor) {
+        // executorId may be a tombstone. Retire every generation mapped to the same actor so a
+        // late kill cannot leave its current generation or actor slot behind.
+        val slotOpt = actorSlots.remove(actorId)
+        val executorIds = slotOpt.map { slot =>
+          Set(executorId) ++ slot.currentExecutorId ++ slot.previousExecutorId
+        }.getOrElse(Set(executorId))
+        executorIds.foreach { id =>
+          removeExecutorGeneration(id)
+          executorIdToActorId.remove(id)
+        }
         // Only explicit Spark/AppMaster shutdown releases the Ray actor slot. A disconnect caused
         // by actor failure keeps the slot so Ray can restart it and register a new executor id.
-        val handlerOpt = actorIdToHandle.remove(exec.actorId)
         // Previously we used to exitExecutor for all scenarios, but it will cause
         // the following issue when a executor is down because of OOM issue:
         // - Executor E1 dies at T0 lets say because of OOm
@@ -154,17 +170,34 @@ private[spark] class ApplicationInfo(
         // - The failed task (stop task) gets retried as there are task retries configured.
         // - The stop task gets fired on the new executor which got recovered
         // - The Recovered executor exits with status as user intended exit.
-        handlerOpt.foreach(handle => RayExecutorUtils.exitExecutor(handle))
+        slotOpt.foreach(slot => RayExecutorUtils.exitExecutor(slot.handle))
+      } else {
+        removeExecutorGeneration(executorId)
+        actorSlots.get(actorId).foreach { slot =>
+          if (slot.currentExecutorId.contains(executorId)) {
+            // Replace the older tombstone so restart history stays bounded to one generation.
+            slot.previousExecutorId.foreach(executorIdToActorId.remove)
+            slot.currentExecutorId = None
+            slot.previousExecutorId = Some(executorId)
+          }
+        }
+        executorIdToActorId(executorId) = actorId
       }
-      true
-    } else {
-      false
+    }
+    actorIdOpt.isDefined
+  }
+
+  private def removeExecutorGeneration(executorId: String): Unit = {
+    executors.remove(executorId).foreach { exec =>
+      removedExecutors += exec
+      coresGranted -= exec.cores
+      exec.address.foreach(addressToExecutorId.remove)
     }
   }
 
   def getExecutorHandler(
       executorId: String): Option[ActorHandle[RayDPExecutor]] = {
-    executorIdToActorId.get(executorId).flatMap(actorIdToHandle.get)
+    executorIdToActorId.get(executorId).flatMap(actorSlots.get).map(_.handle)
   }
 
   def getNextExecutorId(): Int = {
--- a/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
+++ b/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
@@ -19,7 +19,7 @@ package org.apache.spark.deploy.raydp
 
 import java.util.Date
 
-import scala.collection.mutable.{ArrayBuffer, HashMap, HashSet}
+import scala.collection.mutable.{ArrayBuffer, HashMap}
 
 import io.ray.api.ActorHandle
 
@@ -52,17 +52,17 @@ private[spark] class ApplicationInfo(
   var state: ApplicationState.Value = _
   var executors: HashMap[String, ExecutorDesc] = _
   var addressToExecutorId: HashMap[RpcAddress, String] = _
-  var executorIdToHandler: HashMap[String, ActorHandle[RayDPExecutor]] = _
+  var executorIdToActorId: HashMap[String, String] = _
+  var actorIdToHandle: HashMap[String, ActorHandle[RayDPExecutor]] = _
   var removedExecutors: ArrayBuffer[ExecutorDesc] = _
   var coresGranted: Int = _
   var endTime: Long = _
   private var nextExecutorId: Int = _
   // this only count those registered executors and minus removed executors
   private var registeredExecutors: Int = 0
-  // Desired executor target comes from the Spark driver. Actor slots track the Ray actors
-  // AppMaster still owns, independent of transient Spark executor generations.
+  // Desired executor target comes from the Spark driver. Actor handles track the Ray actor
+  // slots AppMaster still owns, independent of transient Spark executor generations.
   private var desiredExecutors: Int = _
-  private var actorSlots: HashSet[String] = _
 
   init()
 
@@ -70,11 +70,11 @@ private[spark] class ApplicationInfo(
     state = ApplicationState.WAITING
     executors = new HashMap[String, ExecutorDesc]
     addressToExecutorId = new HashMap[RpcAddress, String]
-    executorIdToHandler = new HashMap[String, ActorHandle[RayDPExecutor]]
+    executorIdToActorId = new HashMap[String, String]
+    actorIdToHandle = new HashMap[String, ActorHandle[RayDPExecutor]]
     endTime = -1L
     nextExecutorId = 0
     desiredExecutors = desc.numExecutors
-    actorSlots = new HashSet[String]
     removedExecutors = new ArrayBuffer[ExecutorDesc]
   }
 
@@ -86,26 +86,18 @@ private[spark] class ApplicationInfo(
       memoryInMB: Int): Unit = {
     // Adding a pending executor also declares that its Ray actor slot is still active.
     // For restarted executors, actorId points back to the original named Ray actor.
-    actorSlots += actorId
+    actorIdToHandle(actorId) = handler
     val desc = ExecutorDesc(executorId, actorId, cores, memoryInMB, null)
     executors(executorId) = desc
-    executorIdToHandler(executorId) = handler
+    executorIdToActorId(executorId) = actorId
   }
 
   def updateDesiredExecutors(numExecutors: Int): Unit = {
     desiredExecutors = math.max(0, numExecutors)
   }
 
-  def hasActorSlot(actorId: String): Boolean = {
-    actorSlots.contains(actorId)
-  }
-
-  def actorSlotCount: Int = {
-    actorSlots.size
-  }
-
   def numActorsToAdd: Int = {
-    math.max(0, desiredExecutors - actorSlots.size)
+    math.max(0, desiredExecutors - actorIdToHandle.size)
   }
 
   // Compatibility view for ObjectStoreWriter: map restarted Spark executor ids back to
@@ -151,13 +143,14 @@ private[spark] class ApplicationInfo(
       if (exec.registered) {
         registeredExecutors -= 1
       }
-      removedExecutors += executors(executorId)
+      removedExecutors += exec
       executors -= executorId
+      executorIdToActorId -= executorId
       coresGranted -= exec.cores
       if (shutdownActor) {
         // Only explicit Spark/AppMaster shutdown releases the Ray actor slot. A disconnect caused
         // by actor failure keeps the slot so Ray can restart it and register a new executor id.
-        actorSlots -= exec.actorId
+        val handlerOpt = actorIdToHandle.remove(exec.actorId)
         // Previously we used to exitExecutor for all scenarios, but it will cause
         // the following issue when a executor is down because of OOM issue:
         // - Executor E1 dies at T0 lets say because of OOm
@@ -167,9 +160,8 @@ private[spark] class ApplicationInfo(
         // - The failed task (stop task) gets retried as there are task retries configured.
         // - The stop task gets fired on the new executor which got recovered
         // - The Recovered executor exits with status as user intended exit.
-        RayExecutorUtils.exitExecutor(executorIdToHandler(executorId))
+        handlerOpt.foreach(handle => RayExecutorUtils.exitExecutor(handle))
       }
-      executorIdToHandler -= executorId
       true
     } else {
       false
@@ -178,7 +170,7 @@ private[spark] class ApplicationInfo(
 
   def getExecutorHandler(
       executorId: String): Option[ActorHandle[RayDPExecutor]] = {
-    executorIdToHandler.get(executorId)
+    executorIdToActorId.get(executorId).flatMap(actorIdToHandle.get)
   }
 
   def remainingUnRegisteredExecutors(): Int = {
--- a/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
+++ b/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
@@ -183,7 +183,7 @@ private[spark] class ApplicationInfo(
             // - The failed task (stop task) gets retried as there are task retries configured.
             // - The stop task gets fired on the new executor which got recovered
             // - The Recovered executor exits with status as user intended exit.
-            slotOpt.foreach(slot => RayExecutorUtils.exitExecutor(slot.handle))
+            slotOpt.foreach(slot => exitExecutorActor(slot.handle))
             None
         }
       } else {
@@ -208,6 +208,12 @@ private[spark] class ApplicationInfo(
       coresGranted -= exec.cores
       exec.address.foreach(addressToExecutorId.remove)
     }
+  }
+
+  // Visible for testing. Shutting an executor down needs a live Ray actor handle, which unit
+  // tests cannot construct, so the Ray call is isolated behind this method.
+  protected def exitExecutorActor(handle: ActorHandle[RayDPExecutor]): Unit = {
+    RayExecutorUtils.exitExecutor(handle)
   }
 
   def getExecutorHandler(
--- a/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
+++ b/core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
@@ -149,28 +149,43 @@ private[spark] class ApplicationInfo(
 
     actorIdOpt.foreach { actorId =>
       if (shutdownActor) {
-        // executorId may be a tombstone. Retire every generation mapped to the same actor so a
-        // late kill cannot leave its current generation or actor slot behind.
-        val slotOpt = actorSlots.remove(actorId)
-        val executorIds = slotOpt.map { slot =>
-          Set(executorId) ++ slot.currentExecutorId ++ slot.previousExecutorId
-        }.getOrElse(Set(executorId))
-        executorIds.foreach { id =>
-          removeExecutorGeneration(id)
-          executorIdToActorId.remove(id)
+        // One pass over the slot decides whether it survives this kill.
+        actorSlots.updateWith(actorId) {
+          // A tombstone whose actor has already re-registered refers to a generation Spark itself
+          // removed on disconnect. Spark tracks the newer generation as a separate executor, so a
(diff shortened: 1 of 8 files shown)


my-vegetable-has-exploded:
Hi @pang-wu  @carsonwang , could you please review this PR? This fix addresses a real issue we encountered in production.

my-vegetable-has-exploded:
@claude review

pang-wu:
can we add some tests?
Here's a test you can use as a reference to test node kill/restart: https://github.com/ray-project/raydp/blob/master/python/raydp/tests/test_spark_cluster.py#L262

my-vegetable-has-exploded:
> can we add some tests? Here's a test you can use as a reference to test node kill/restart: https://github.com/ray-project/raydp/blob/master/python/raydp/tests/test_spark_cluster.py#L262

I have added e2e tests. Those tests fail in pre-fix backend but pass in fixed backend.

```
___________ test_kill_and_replace_executor_reconciles_desired_count ____________

jdk17_extra_spark_configs = {'spark.executor.extraJavaOptions': '-XX:+IgnoreUnrecognizedVMOptions --add-opens=java.base/java.lang=ALL-UNNAMED --ad...ALL-UNNAMED --add-opens=java.base/sun.security.action=ALL-UNNAMED --add-opens=java.base/sun.util.calendar=ALL-UNNAMED'}

    def test_kill_and_replace_executor_reconciles_desired_count(jdk17_extra_spark_configs):
        """Kill one of two executors and verify that a different executor restores the count to two.
    
        Before the fix, the backend removes the victim and stays at one. After the fix, reconciliation
        creates a replacement and returns to the original desired count of two.
        """
        cluster = Cluster(initialize_head=True, head_node_args={"num_cpus": 1})
        spark = None
        try:
            ray.init(address=cluster.address, include_dashboard=False)
            # Provide exactly two executor slots. Killing one frees all resources needed to replace it.
            cluster.add_node(
                num_cpus=2,
                object_store_memory=10 ** 8,
                resources={"spark_executor": 2},
            )
            configs = {
                "spark.ray.raydp_spark_executor.actor.resource.spark_executor": "1",
                **jdk17_extra_spark_configs,
            }
            spark = raydp.init_spark(
                "test_kill_and_replace", 2, 1, "500M", configs=configs)
            jsc = spark.sparkContext._jsc.sc()
    
            def executor_ids():
                ids = jsc.getExecutorIds()
                return {ids.apply(i) for i in range(ids.size())}
    
            # Verify that both requested executors are registered, then select one ID to replace.
            initial_ids = _wait_for(executor_ids, lambda ids: len(ids) == 2)
            victim_id = next(iter(initial_ids))
    
            # Assert that Spark accepts the kill-and-replace request. The following wait requires both
            # count == 2 and victim removal: base times out at count 1; HEAD returns with a new ID.
            assert jsc.killAndReplaceExecutor(victim_id)
>           replacement_ids = _wait_for(
                executor_ids,
                lambda ids: len(ids) == 2 and victim_id not in ids,
            )

/workspace/python/raydp/tests/test_spark_cluster.py:502: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

value_fn = <function test_kill_and_replace_executor_reconciles_desired_count.<locals>.executor_ids at 0x7f2455d84a60>
condition = <function test_kill_and_replace_executor_reconciles_desired_count.<locals>.<lambda> at 0x7f2455d84b80>
timeout = 60

    def _wait_for(value_fn, condition, timeout=60):
        deadline = time.monotonic() + timeout
        value = None
        while time.monotonic() < deadline:
            value = value_fn()
            if condition(value):
                return value
            time.sleep(0.5)
>       raise AssertionError(f"Condition not met within {timeout}s; last value: {value!r}")
E       AssertionError: Condition not met within 60s; last value: {'0'}

/workspace/python/raydp/tests/test_spark_cluster.py:44: AssertionError
----------------------------- Captured stdout call -----------------------------
2026-08-16 22:12:17,684 ERROR TaskSchedulerImpl [dispatcher-CoarseGrainedScheduler]: Lost executor 1 on 172.17.0.11: Remote RPC client disassociated. Likely due to containers exceeding thresholds, or network issues. Check driver logs for WARN messages.
----------------------------- Captured stderr call -----------------------------
2026-08-16 22:12:02,278 INFO worker.py:1814 -- Connecting to existing Ray cluster at address: 172.17.0.11:38715...
2026-08-16 22:12:02,294 INFO worker.py:2003 -- Connected to Ray cluster. View the dashboard at 127.0.0.1:8265 
(RayDPSparkMaster pid=4927) [2026-08-16 22:12:07,740 I 4988 5093] gcs_client.cc:96: GcsClient has no Cluster ID set, and won't fetch from GCS.
(RayDPSparkMaster pid=4927) [2026-08-16 22:12:07,931 I 4988 5093] gcs_client.cc:96: GcsClient has no Cluster ID set, and won't fetch from GCS.
(raylet) [2026-08-16 22:12:14,830 I 5555 5600] gcs_client.cc:96: GcsClient has no Cluster ID set, and won't fetch from GCS. [repeated 2x across cluster]
(raylet) [2026-08-16 22:12:14,929 I 5559 5601] gcs_client.cc:96: GcsClient has no Cluster ID set, and won't fetch from GCS.
------------------------------ Captured log call -------------------------------
2026-08-16 22:12:02,278 INFO worker.py:1814 -- Connecting to existing Ray cluster at address: 172.17.0.11:38715...
2026-08-16 22:12:02,294 INFO worker.py:2003 -- Connected to Ray cluster. View the dashboard at 127.0.0.1:8265 
=============================== warnings summary ===============================
raydp/tests/test_spark_cluster.py::test_restarted_executor_does_not_block_scale_up
  /opt/miniconda3/lib/python3.10/site-packages/ray/_private/worker.py:2051: FutureWarning: Tip: In future versions of Ray, Ray will no longer override accelerator visible devices env var if num_gpus=0 or num_gpus=None (default). To enable this behavior and turn off this error message, set RAY_ACCEL_ENV_VAR_OVERRIDE_ON_ZERO=0
    warnings.warn(

../opt/miniconda3/lib/python3.10/site-packages/_pytest/cacheprovider.py:469
  /opt/miniconda3/lib/python3.10/site-packages/_pytest/cacheprovider.py:469: PytestCacheWarning: could not create cache path /workspace/python/.pytest_cache/v/cache/nodeids: [Errno 30] Read-only file system: '/workspace/python/pytest-cache-files-bqbfs3tu'
    config.cache.set("cache/nodeids", sorted(self.cached_nodeids))

../opt/miniconda3/lib/python3.10/site-packages/_pytest/cacheprovider.py:423
  /opt/miniconda3/lib/python3.10/site-packages/_pytest/cacheprovider.py:423: PytestCacheWarning: could not create cache path /workspace/python/.pytest_cache/v/cache/lastfailed: [Errno 30] Read-only file system: '/workspace/python/pytest-cache-files-po6nlvhd'
    config.cache.set("cache/lastfailed", self.lastfailed)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
FAILED ../workspace/python/raydp/tests/test_spark_cluster.py::test_restarted_executor_does_not_block_scale_up - AssertionError: Condition not met within 60s; last value: {'4', '0'}
FAILED ../workspace/python/raydp/tests/test_spark_cluster.py::test_kill_and_replace_executor_reconciles_desired_count - AssertionError: Condition not met within 60s; last value: {'0'}
================== 2 failed, 3 warnings in 178.03s (0:02:58) ===================
```

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
```diff
@@ -188,25 +202,25 @@ class RayAppMaster(host: String,
...
-            val handler = handlerOpt.get.asInstanceOf[ActorHandle[RayDPExecutor]]
-            appInfo.addPendingRegisterExecutor(newExecutorId, handler, cores, memory)
-            restartedExecutors(newExecutorId) = executorId
-            context.reply(AddPendingRestartedExecutorReply(Some(newExecutorId)))
-          }
-        } else {
+      case RequestAddPendingRestartedExecutor(actorId) =>
+        val cores = appInfo.desc.coresPerExecutor.getOrElse(1)
+        val memory = appInfo.desc.memoryPerExecutorMB
+        // ray actor will restart using the old ID
```
pang-wu, line 204: **Blocking: the failure edge isn't reconciled, only the deliberate-removal edge is.**

Reconcile is added here after `KillExecutors`, but `onDisconnected` (`kill(shutdownActor = false)`) keeps the slot and does *not* reconcile, with no timeout or liveness check anywhere.

Retaining the slot on disconnect is correct only if `setMaxRestarts(-1)` guarantees the actor eventually returns. When it can't — placement-group bundle gone, node permanently evicted — the retained slot becomes a permanent ghost: `numActorsToAdd = desiredExecutors - actorIdToHandle.size` sees it as present, so no replacement is created, and no `KillExecutors` arrives (Spark won't resend the unchanged total, per this PR's own YARN argument). The app then runs below `desiredExecutors` indefinitely — the mirror image of the double-count bug this PR fixes.

Neither new test covers this negative path. Needs a bounded slot retention (retire slot + reconcile if no re-registration within a timeout), a liveness probe of the named actor, or finite `maxRestarts` with an explicit give-up→replace path.

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
```diff
@@ -49,34 +52,61 @@ private[spark] class ApplicationInfo(
...
+  }
+
+  def numActorsToAdd: Int = {
+    math.max(0, desiredExecutors - actorIdToHandle.size)
+  }
+
+  // Compatibility view for ObjectStoreWriter: map restarted Spark executor ids back to
+  // the original Ray actor ids used in named actor lookup.
+  def getRestartedExecutors: Map[String, String] = {
+    executors.collect {
```
pang-wu, line 115: **Unsynchronized cross-thread read of a mutable map.**

`getRestartedExecutors` is reachable from `RayAppMaster.getRestartedExecutors()`, which is invoked as a **Ray actor task** (a different thread) while `receive`/`receiveAndReply` mutate `executors` on Spark's **RPC dispatcher thread**. Iterating `executors` (a `scala.collection.mutable.HashMap`) concurrently with writes can throw or return corrupt data.

The old code read a separate `restartedExecutors` map; deriving from the live `executors` map widens the shared write surface, so the race is slightly worse now. Suggest guarding the read (and the mutating paths) with a lock, or returning a defensive snapshot taken on the RPC thread.

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
```diff
@@ -49,34 +52,61 @@ private[spark] class ApplicationInfo(
...
-    executorIdToHandler(executorId) = handler
+    executorIdToActorId(executorId) = actorId
+  }
+
+  def updateDesiredExecutors(numExecutors: Int): Unit = {
+    desiredExecutors = math.max(0, numExecutors)
+  }
+
+  def numActorsToAdd: Int = {
+    math.max(0, desiredExecutors - actorIdToHandle.size)
```
pang-wu: After this PR, `currentExecutors()` and `remainingUnRegisteredExecutors()` have no callers (both former call sites in `RequestExecutors` and `RequestAddPendingRestartedExecutor` were removed), and `registeredExecutors` now only feeds those two dead methods. Either delete them, or — better — repurpose `registeredExecutors` as the liveness signal for the `onDisconnected` gap noted on the RayAppMaster side.

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
```diff
@@ -250,8 +265,13 @@ class RayAppMaster(host: String,
...
     private def schedule(): Unit = {
-      val desc = appInfo.desc
-      for (_ <- 0 until desc.numExecutors) {
+      reconcileExecutors()
+    }
+
+    private def reconcileExecutors(): Unit = {
+      // Reconcile against Ray actor slots, not registered Spark executor generations.
+      // Restarted actors keep the same slot and only receive a new Spark executor id.
+      (0 until appInfo.numActorsToAdd).foreach { _ =>
```
pang-wu, line 269: Minor: the old `spark.dynamicAllocation.maxExecutors` / `spark.executor.instances` hard-cap backstop in `requestNewExecutor` is now gone, so reconcile-to-`desiredExecutors` is the only bound on actor creation. That's defensible since `desiredExecutors` is driven by Spark's requested total, but it removes the last defensive clamp — any future bug in `desiredExecutors` tracking would provision unbounded actors instead of being capped. Worth a one-line comment stating the cap is intentionally delegated to `desiredExecutors`.

pang-wu (commented):
Latest revision looks like a real improvement over earlier ones — the actor-slot model is the correct root-cause fix, and the new e2e tests plus the parameter rename tighten it up. I have one blocking concern (the `onDisconnected` failure edge) and three smaller notes, left inline. Nothing here disputes the core design; it's about closing the remaining failure path and some cleanup.

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
my-vegetable-has-exploded: The previous implementation called getRestartedExecutors directly on the RayAppMasterEndpoint instance, which bypassed the Spark RPC endpoint mailbox and could read appInfo concurrently with updates made in receiveAndReply.
I changed the call to send a GetRestartedExecutors message through RpcEndpointRef.askSync. The read is now handled inside receiveAndReply and serialized with all other appInfo accesses by ThreadSafeRpcEndpoint.

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
my-vegetable-has-exploded: done！

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
my-vegetable-has-exploded: Thanks for review @pang-wu.

> When it can't — placement-group bundle gone, node permanently evicted — the retained slot becomes a permanent ghost

1、If only the Bundle node becomes unavailable, the restart strategy will schedule the actor onto a new node.
2、As far as I know , this fix mainly consists of two parts: liveness detection and recovery. But there is a question: what exactly does the recovery strategy need to do? Among the cases I can currently think of, if rebuilding the actor can solve the problem, the restart strategy seems to be able to solve it as well.

pang-wu: Agreed on the bundle node — with `setMaxRestarts(-1)` Ray reschedules onto another node. And where the cluster genuinely cannot satisfy the bundle, a replacement actor would sit pending just as the original does, so liveness detection plus recreation adds cost without adding recovery. This is not a liveness problem.

On "if rebuilding the actor can solve the problem, the restart strategy can solve it as well": restart is not universal recovery. `RayDPExecutor.stop()` is `Ray.exitActor`, an **intentional** exit, which Ray does not restart regardless of `maxRestarts(-1)`. Two exit classes, different behavior — which is why slot bookkeeping cannot fully delegate to Ray's restart strategy.

## The gap is bookkeeping

It hinges on one line: `ApplicationInfo.kill()` early-returns unless `executors.contains(executorId)`. Releasing a slot is therefore keyed on a Spark executor *generation* that may already be gone.

**A. `KillExecutors` arrives while the generation is alive**

```
  Spark Driver              RayAppMaster                     Ray / actor 23
       |                          |                                |
       |  KillExecutors(["1001"]) |                                |
       |------------------------->|                                |
       |                          | kill("1001", shutdown=true)    |
       |                          |  |- executors.contains OK      |
       |                          |  |- executors       -= 1001    |
       |                          |  |- actorIdToHandle -= 23  <-- SLOT RELEASED
       |                          |  \- exitExecutor ------------->| Ray.exitActor
       |                          |                                | (intentional ->
       |<-------- true -----------|                                |  NO restart)
       |                          | reconcileExecutors()           |
       |                          |  numActorsToAdd = 1 -> new actor
```

Slot released, reconcile sees the deficit, replacement created. Correct.

**B. The kill lands after the generation already left `executors`**

```
  Spark Driver              RayAppMaster                     Ray / actor 23
       |                          |                                |
       |  (kill for 1001 already  |                                |
       |   decided and in flight) |                                | X JVM dies
       |                          |<--- conn drop -----------------|
       |                          | onDisconnected(addrX)          |
       |                          | kill("1001", shutdown=false)   |
       |                          |  |- executors           -= 1001|
       |                          |  |- executorIdToActorId -= 1001|
       |                          |  \- actorIdToHandle KEEPS 23 <-- slot retained
       |                          |     (deliberate: Ray will restart)
       |                          |                                | Ray begins restart
       |  KillExecutors(["1001"]) |                                |
       |------------------------->|                                |
       |                          | kill("1001", shutdown=true)    |
       |                          |  \- executors.contains FAILS   |
       |                          |     -> return false, NO-OP <---- slot NOT released
       |<------- false -----------|                                |  exitActor NOT sent
       |                          | reconcileExecutors()           |
       |                          |  slot 23 still counted         |
       |                          |  -> numActorsToAdd = 0, no repl|
       |                          |                                | restart completes
       |                          |<-- RequestAddPending(23) ------|
       |                          |  actorIdToHandle.contains(23)? |
       |                          |  -> TRUE (never removed)       |
       |                          |  -> new executorId "1002"      |
       |                          |--- reply Some("1002") -------->|
       |<---- executor 1002 registers ------------------------------|
```

State side by side:

| after | `executors` | `executorIdToActorId` | `actorIdToHandle` (slots) | actor 23 |
| --- | --- | --- | --- | --- |
| start | `{1001}` | `{1001->23}` | `{23->h}` | alive |
| **A** kill | `{}` | `{}` | `{}` | exited, no restart |
| **B** disconnect | `{}` | `{}` | `{23->h}` | restarting |
| **B** kill (no-op) | `{}` | `{}` | `{23->h}` | restarting |
| **B** re-register | `{1002}` | `{1002->23}` | `{23->h}` | **alive again** |

In **B**, Spark explicitly asked for that executor to go away and it came back as `1002`. The `!actorIdToHandle.contains(actorId)` guard — whose stated purpose is *"Do not allow a late restart request to recreate a removed executor"* — cannot fire, because the slot was never removed. `doKillExecutors` also returned `false` to the driver.

## What ordering B requires

Two facts bound it:

- `ExecutorStarted` is sent from `env.rpcEnv`, the executor's real SparkEnv RpcEnv — the same process and listening RpcEnv the driver is attached to. The address AppMaster records in `addressToExecutorId` belongs to that same RpcEnv, so a process death closes both peers' connections from the same event.
- Spark's `CoarseGrainedSchedulerBackend.killExecutors` partitions requested ids against `executorDataMap` and drops unknown ones before calling `doKillExecutors`. Once the driver has processed its own disconnect, it will not send `KillExecutors` for that executor at all.

So **B** requires the kill to already be in flight — decided and sent before the driver processed its own disconnect — and to arrive at the AppMaster inbox after `OnDisconnected`. Both messages land in the same endpoint inbox and are processed serially in arrival order, so there is no mailbox reordering to exploit; it comes down to arrival order within roughly one RTT. A genuine ordering hazard, but a narrow window rather than a routine path.

The failure is silent: nothing logs it, nothing reconciles it, `doKillExecutors` reports `false`, and no path retries. The slot stays occupied for the remaining life of the application while Spark believes that executor is gone.

## Suggested fix

Two changes, both in `ApplicationInfo`.

**1. `kill()` — resolve the slot through `executorIdToActorId`, so a kill still lands when the generation is already gone:**

```diff
   def kill(executorId: String, shutdownActor: Boolean): Boolean = {
-    if (executors.contains(executorId)) {
-      val exec = executors(executorId)
-      if (exec.registered) {
-        registeredExecutors -= 1
-      }
-      removedExecutors += exec
-      executors -= executorId
-      executorIdToActorId -= executorId
-      coresGranted -= exec.cores
-      if (shutdownActor) {
-        // Only explicit Spark/AppMaster shutdown releases the Ray actor slot. A disconnect caused
-        // by actor failure keeps the slot so Ray can restart it and register a new executor id.
-        val handlerOpt = actorIdToHandle.remove(exec.actorId)
-        // Previously we used to exitExecutor for all scenarios, but it will cause
-        // the following issue when a executor is down because of OOM issue:
-        // - Executor E1 dies at T0 lets say because of OOm
-        // - We try to kill it by firing stop call on E1 actor
-        // - Since the actor is not available, the stop task fails for E1
-        // - In the mean while, ray brings up the lost executor E1
-        // - The failed task (stop task) gets retried as there are task retries configured.
-        // - The stop task gets fired on the new executor which got recovered
-        // - The Recovered executor exits with status as user intended exit.
-        handlerOpt.foreach(handle => RayExecutorUtils.exitExecutor(handle))
-      }
-      true
-    } else {
-      false
+    val execOpt = executors.remove(executorId)
+    execOpt.foreach { exec =>
+      if (exec.registered) {
+        registeredExecutors -= 1
+      }
+      removedExecutors += exec
+      coresGranted -= exec.cores
+    }
+
+    // Resolve the stable actor id even when this Spark executor generation was already
+    // dropped by an earlier onDisconnected, so a later kill can still find the slot.
+    val actorIdOpt = execOpt.map(_.actorId).orElse(executorIdToActorId.get(executorId))
+
+    if (shutdownActor) {
+      // Only explicit Spark/AppMaster shutdown releases the Ray actor slot. A disconnect caused
+      // by actor failure keeps the slot so Ray can restart it and register a new executor id.
+      actorIdOpt.foreach { actorId =>
+        executorIdToActorId -= executorId
+        val handlerOpt = actorIdToHandle.remove(actorId)
+        // Previously we used to exitExecutor for all scenarios, but it will cause
+        // the following issue when a executor is down because of OOM issue:
+        // - Executor E1 dies at T0 lets say because of OOm
+        // - We try to kill it by firing stop call on E1 actor
+        // - Since the actor is not available, the stop task fails for E1
+        // - In the mean while, ray brings up the lost executor E1
+        // - The failed task (stop task) gets retried as there are task retries configured.
+        // - The stop task gets fired on the new executor which got recovered
+        // - The Recovered executor exits with status as user intended exit.
+        handlerOpt.foreach(handle => RayExecutorUtils.exitExecutor(handle))
+      }
     }
+    // On disconnect the executorId -> actorId entry is deliberately retained as a tombstone,
+    // so a later KillExecutors for that generation can still resolve and release the slot.
+    actorIdOpt.isDefined
   }
```

**2. `addPendingRegisterExecutor()` — bound the tombstones so they cannot accumulate:**

```diff
   def addPendingRegisterExecutor(
       executorId: String,
       actorId: String,
       handler: ActorHandle[RayDPExecutor],
       cores: Int,
       memoryInMB: Int): Unit = {
     // Adding a pending executor also declares that its Ray actor slot is still active.
     // For restarted executors, actorId points back to the original named Ray actor.
     actorIdToHandle(actorId) = handler
+    // Drop tombstones left by earlier generations of this actor. Once it has re-registered
+    // under a new executor id, stale ids must no longer resolve to this slot.
+    executorIdToActorId.filterInPlace { case (_, mapped) => mapped != actorId }
     val desc = ExecutorDesc(executorId, actorId, cores, memoryInMB, null)
     executors(executorId) = desc
     executorIdToActorId(executorId) = actorId
   }
```

With (2) at most one stale `executorId` points at a given `actorId` at any time, and it is cleared the moment that actor registers its next generation. The tradeoff is that a kill arriving after the next generation has registered no longer resolves to the slot, which is the intended behaviour — Spark is tracking the new executor id by then.

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
pang-wu: Still open on the latest revision (`aaf068b`) — `currentExecutors()` and `remainingUnRegisteredExecutors()` are both still present with zero callers repo-wide, and `registeredExecutors` is now maintained solely to feed them.

One small related item from the `GetRestartedExecutors` change: `appMasterEndpoint` is now only used as the argument to `setupEndpoint` on the very next line, since `getRestartedExecutors()` goes through the `endpoint` ref instead. It can just be a local `val` in `init()` rather than a field.

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
pang-wu: This race is pre-existing rather than something this PR introduces — the old code had the same shape, with `restartedExecutors` written by the RPC handler and read from the actor-task path. The change here only shifts the read onto the `executors` map.

Resolved by the `GetRestartedExecutors` message plus `askSync` in `aaf068b`, and more cleanly than the lock suggested above: routing through the endpoint mailbox serializes the read with every other `appInfo` access and removes the direct-instance call that bypassed the mailbox entirely.

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/RayAppMaster.scala
my-vegetable-has-exploded: Thanks for pointing this out @pang-wu . I fixed the race between executor disconnect and a delayed KillExecutors message.
After a disconnect, the previous executor ID is now retained as a tombstone and continues to resolve to the same Ray actor slot. If a delayed kill arrives for that ID after the actor has restarted, RayDP removes the slot together with both the previous and current executor generations, then shuts down the actor. Only the latest previous executor ID is retained, so the state remains bounded.

core/raydp-main/src/main/scala/org/apache/spark/deploy/raydp/ApplicationInfo.scala
my-vegetable-has-exploded: done.

my-vegetable-has-exploded:
I opened #488 to resovle ci problem.

pang-wu (approved):
lgtm -- @my-vegetable-has-exploded I patched your branch adding additional tests and style fixes

my-vegetable-has-exploded:
Thanks!!! @pang-wu

PR #488, opened 2026-09-03
ci: pin Google API protos for TensorFlow compatibility

my-vegetable-has-exploded:
## Summary

- Pin `googleapis-common-protos` to 1.75.0, which remains compatible with the protobuf 4.x runtime installed by TensorFlow 2.15.
- Add a fast `google.rpc.code_pb2` import check before lint and pytest.
- Fix three `Google Collab` misspellings in the README.

## Background

`googleapis-common-protos` 1.75.2 requires protobuf 6.33.5 or newer, while the RayDP TensorFlow extra installs TensorFlow 2.15.1 and protobuf 4.25.9. The generated Google RPC modules then fail to import `google.protobuf.runtime_version`, causing the Ray dashboard agent and raylet to exit.

### DIFF
--- a/.github/workflows/raydp.yml
+++ b/.github/workflows/raydp.yml
@@ -82,7 +82,9 @@ jobs:
           else
             pip install torch
           fi
-          pip install pyarrow "ray[train,default]==${{ matrix.ray-version }}" tqdm pytest tabulate grpcio-tools wget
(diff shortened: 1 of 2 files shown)


my-vegetable-has-exploded:
@pang-wu @carsonwang ptal，origin ci is broken.
image

pang-wu (approved):
lgtm, thanks for the fix

PR #490, opened 2026-09-06
Support every Spark 4.x patch release and add Spark 4.2

pang-wu:
## What does this PR do?

Two related changes to Spark version support:

1. **Shims now cover a whole `major.minor` line**, so every patch release of a supported minor works without a code change.
2. **Adds a shim for Spark 4.2** and registers `4.2.0` in both test matrices.

## Why is this change needed?

Each shim provider enumerated the patch releases it claimed, and `matches()` compared the running Spark version against that list by exact string:

```scala
private val SUPPORTED_PATCHES = 0 to 1
val DESCRIPTOR_STRINGS = SUPPORTED_PATCHES.map(p => SparkShimDescriptor(4, 1, p).toString)
def matches(version: String): Boolean = DESCRIPTOR_STRINGS.contains(version)
```

Spark keeps patch releases source and binary compatible within a minor line, but any patch outside the hard-coded range fails shim resolution at runtime with `No Spark Shim Provider found for <version>`. Several released versions were already unreachable:

| Line | Released patches | Previously claimed |
| ---- | ---------------- | ------------------ |
| 4.0.x | 4.0.0 – 4.0.4 | 4.0.0 – 4.0.2 |
| 4.1.x | 4.1.0 – 4.1.3 | 4.1.0 – 4.1.1 |

So `pyspark==4.1.3` failed even though nothing in the shim was incompatible, and every future patch would need a follow-up PR.

`SparkMinorLineShimProvider` now parses the running version and claims any patch of the line it was constructed for. Two details are deliberate:

- Parsing tolerates a qualifier, so `4.2.0-preview1` and `4.3.0-SNAPSHOT` resolve to their base version.
- Matching compares parsed components rather than string prefixes, so the `4.1` line does not claim `4.10.x`.

The `pyspark` requirement becomes a bounded range over the supported lines (`>= 4.0.0, < 4.3.0`) instead of a specific patch ceiling, so pip can install any patch we cover and nothing past it.

The 3.x shims are left alone — they enumerate patch lists (and `spark322` deliberately claims some 3.1.x versions), so converting them is a separate decision with its own compatibility question.

## Related Issue

Fixes #

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Test update
- [ ] Refactoring
- [ ] Other

## How was this tested?

**New unit tests** — `core/raydp-main/src/test/scala/com/intel/raydp/shims/SparkShimDescriptorTest.scala`, 7 cases covering version parsing, qualifier suffixes, unparseable input, every patch of a line matching, other minor lines not matching, and the `4.1` vs `4.10` boundary.

```
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0 -- SparkShimDescriptorTest
Tests run: 7, Failures: 0, Errors: 0, Skipped: 0 -- ApplicationInfoTest
Tests run: 14, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

**Full build of all 12 modules**, including the new shim compiling against Spark 4.2.0:

```
RayDP Shims for Spark 4.0.0 ........................ SUCCESS
RayDP Shims for Spark 4.1.0 ........................ SUCCESS
RayDP Shims for Spark 4.2 .......................... SUCCESS
BUILD SUCCESS
```

The Spark 4.2 shim needed no source changes beyond the package and class renames, so the shimmed APIs are unchanged between 4.1 and 4.2.

Also verified that the new shim jar is named `raydp-shims-spark420-1.7.0-SNAPSHOT.jar`, which matches the `raydp-*.jar` glob `python/setup.py` uses to bundle jars into the wheel, and that no new lines exceed the 100-column scalastyle limit.

**Not run locally:** the pytest suite, which needs a live Ray cluster, and any run against a real 4.2.0 cluster. CI covers those across the matrix.

## Checklist

- [x] I have added or updated tests if needed
- [ ] I have updated documentation if needed
- [x] I have run relevant tests locally
- [x] I have checked that this change is backward compatible

### DIFF
--- a/.github/workflows/ray_nightly_test.yml
+++ b/.github/workflows/ray_nightly_test.yml
@@ -32,7 +32,7 @@ jobs:
       matrix:
         os: [ ubuntu-latest ]
         python-version: ["3.10", "3.11"]
-        spark-version: [4.0.0, 4.1.0]
+        spark-version: [4.0.0, 4.1.0, 4.2.0]
 
     runs-on: ${{ matrix.os }}
 
--- a/.github/workflows/raydp.yml
+++ b/.github/workflows/raydp.yml
@@ -33,7 +33,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: ["3.10", "3.11"]
-        spark-version: [4.0.0, 4.1.0]
+        spark-version: [4.0.0, 4.1.0, 4.2.0]
         ray-version: [2.37.0, 2.40.0, 2.50.0]
 
     runs-on: ${{ matrix.os }}
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -20,6 +20,7 @@
     <spark350.version>3.5.0</spark350.version>
     <spark400.version>4.0.0</spark400.version>
     <spark410.version>4.1.0</spark410.version>
+    <spark420.version>4.2.0</spark420.version>
     <snappy.version>1.1.10.4</snappy.version>
     <netty.version>4.1.94.Final</netty.version>
     <commons.text.version>1.10.0</commons.text.version>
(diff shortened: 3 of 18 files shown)


carsonwang (approved):
Thanks for the work.

